PR #94, opened 2022-10-26
Clean up github workflows

hasheddan:

Drops updoc from generation to ensure we don't need to depend on a
private package.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

Drops the pull_request_target trigger and only uses pull_request. Also
drops the private repo dependency and conditionally publishes on access
to push secrets.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

I have:

Will need to verify on this PR and post-merge.


ulucinar (approved):
Thank you @hasheddan, lgtm.

PR #139, opened 2022-11-10
Make not-generated Terraform resources CSV available as a Github workflow artifact

ulucinar:

This PR proposes a change where a CSV file of Terraform resource names which are available in the Terraform provider's schema but not generated as managed resources is produced and made available as a Github workflow artifact.

Currently, the artifact is generated when the `generate` make target is run as a dependency (of `check-diff`). This is a little bit implicit and as an alternative we can generate the CSV on its own step. But I did not want to add to our build times (by running `make generate` a second time or computing the diff from the JSON schema and the generated files), and thus preferred to generate the CSV in the `Check Diff` step. 

I have:

Tested with the following workflow run:
https://github.com/upbound/provider-aws/actions/runs/3438916965

### DIFF
--- a/cmd/generator/main.go
+++ b/cmd/generator/main.go
@@ -1,6 +1,16 @@
-/*
-Copyright 2021 Upbound Inc.
-*/
+// Copyright 2022 Upbound Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
 
 package main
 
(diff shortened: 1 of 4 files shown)


cmd/generator/main.go
```diff
@@ -23,5 +24,11 @@ func main() {
 	if err != nil {
 		panic(fmt.Sprintf("cannot calculate the absolute path with %s", rootDir))
 	}
-	pipeline.Run(config.GetProvider(), absRootDir)
+	p := config.GetProvider()
+	pipeline.Run(p, absRootDir)
+	if fp := os.Getenv("SKIPPED_RESOURCES_CSV"); len(fp) != 0 {
+		if err := os.WriteFile(fp, []byte(strings.Join(p.GetSkippedResourceNames(), ",")), 0o600); err != nil {
```
jbw976, line 30: should these be using `;` so each resource is on its own row? looks like all the resources are generated into a single row now.  Not sure which is better, it's just the first thing I noticed when opening the csv is that you have to scroll horizontally forever because they are in one row 😇

ulucinar: Replaced the separator char with `;`. @jbw976, which spreadsheet editor are you using? After seeing your comment, I gave it a try and it's opened by Numbers. With Numbers, both `,` and `;` result in a table with a single row. If we would prefer a table with a single column, then we may use `\n` as the separator char here.

cmd/generator/main.go
```diff
@@ -23,5 +24,11 @@ func main() {
 	if err != nil {
 		panic(fmt.Sprintf("cannot calculate the absolute path with %s", rootDir))
 	}
-	pipeline.Run(config.GetProvider(), absRootDir)
+	p := config.GetProvider()
+	pipeline.Run(p, absRootDir)
+	if fp := os.Getenv("SKIPPED_RESOURCES_CSV"); len(fp) != 0 {
```
turkenh, line 29: You may consider using a cli parsing library and accept both flags and env vars to keep argument parsing consistent like in [uptest](https://github.com/upbound/uptest/blob/581f891e0a73d6d9dd236a66e07f72c5047d9ad3/cmd/main.go#L23).

But it is totally up to you to deal with it now or later if we need more inputs. But at least a comment on why we use env here would be helpful.

turkenh (approved):

cmd/generator/main.go
ulucinar: Done. Moved to the `kingpin.v2` package for argument parsing in the generator.

cmd/generator/main.go
jbw976: yes sir, Pages gets opened by default on my Mac.  When I mentioned `;`, I didn't try it out or think it through :)

My personal opinion is that whatever works to make it a single column instead of single row is a more friendly scrolling experience.  But that is not a super strong opinion and you can make the call here 🙇

ulucinar: Then, let's proceed with a `\n` as the separator char to get a single column table.

PR #140, opened 2022-11-10
Configurable maxConcurrentReconciles and pollInterval with sensible defaults

turkenh:

This PR makes maxConcurrentReconciles and pollInterval configurable with the proposed defaults experimented [here](https://github.com/upbound/upjet/issues/116#issuecomment-1308941093).
It follows the same approach we used in the community providers to use input arguments as internal configurations.

Fixes https://github.com/upbound/upjet/issues/116

I have:

See https://github.com/upbound/upjet/issues/116#issuecomment-1308941093

### DIFF
--- a/cmd/provider/main.go
+++ b/cmd/provider/main.go
@@ -37,13 +37,14 @@ func main() {
 	var (
 		app                = kingpin.New(filepath.Base(os.Args[0]), "AWS support for Crossplane.").DefaultEnvars()
 		debug              = app.Flag("debug", "Run with debug logging.").Short('d').Bool()
-		syncPeriod         = app.Flag("sync", "Controller manager sync period such as 300ms, 1.5h, or 2h45m").Short('s').Default("1h").Duration()
+		syncInterval       = app.Flag("sync", "Sync interval controls how often all resources will be double checked for drift.").Short('s').Default("1h").Duration()
+		pollInterval       = app.Flag("poll", "Poll interval controls how often an individual resource should be checked for drift.").Default("10m").Duration()
 		leaderElection     = app.Flag("leader-election", "Use leader election for the controller manager.").Short('l').Default("false").OverrideDefaultFromEnvar("LEADER_ELECTION").Bool()
+		maxReconcileRate   = app.Flag("max-reconcile-rate", "The global maximum rate per second at which resources may checked for drift from the desired state.").Default("10").Int()
 		terraformVersion   = app.Flag("terraform-version", "Terraform version.").Required().Envar("TERRAFORM_VERSION").String()
 		providerSource     = app.Flag("terraform-provider-source", "Terraform provider source.").Required().Envar("TERRAFORM_PROVIDER_SOURCE").String()
 		providerVersion    = app.Flag("terraform-provider-version", "Terraform provider version.").Required().Envar("TERRAFORM_PROVIDER_VERSION").String()
 		nativeProviderPath = app.Flag("terraform-native-provider-path", "Terraform native provider path for shared execution.").Default("").Envar("TERRAFORM_NATIVE_PROVIDER_PATH").String()
-		maxReconcileRate   = app.Flag("max-reconcile-rate", "The global maximum rate per second at which resources may checked for drift from the desired state.").Default("10").Int()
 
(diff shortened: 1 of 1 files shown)


cmd/provider/main.go
```diff
@@ -37,13 +37,14 @@ func main() {
 	var (
 		app                = kingpin.New(filepath.Base(os.Args[0]), "AWS support for Crossplane.").DefaultEnvars()
 		debug              = app.Flag("debug", "Run with debug logging.").Short('d').Bool()
-		syncPeriod         = app.Flag("sync", "Controller manager sync period such as 300ms, 1.5h, or 2h45m").Short('s').Default("1h").Duration()
+		syncInterval       = app.Flag("sync", "Sync interval controls how often all resources will be double checked for drift.").Short('s').Default("1h").Duration()
+		pollInterval       = app.Flag("poll", "Poll interval controls how often an individual resource should be checked for drift.").Default("10m").Duration()
 		leaderElection     = app.Flag("leader-election", "Use leader election for the controller manager.").Short('l').Default("false").OverrideDefaultFromEnvar("LEADER_ELECTION").Bool()
+		maxReconcileRate   = app.Flag("max-reconcile-rate", "The global maximum rate per second at which resources may checked for drift from the desired state.").Default("10").Int()
```
ulucinar, line 43:
```suggestion
		maxReconcileRate   = app.Flag("max-reconcile-rate", "The global maximum rate per second at which resources may be checked for drift from the desired state.").Default("10").Int()
```

cmd/provider/main.go
```diff
@@ -37,13 +37,14 @@ func main() {
 	var (
 		app                = kingpin.New(filepath.Base(os.Args[0]), "AWS support for Crossplane.").DefaultEnvars()
 		debug              = app.Flag("debug", "Run with debug logging.").Short('d').Bool()
-		syncPeriod         = app.Flag("sync", "Controller manager sync period such as 300ms, 1.5h, or 2h45m").Short('s').Default("1h").Duration()
+		syncInterval       = app.Flag("sync", "Sync interval controls how often all resources will be double checked for drift.").Short('s').Default("1h").Duration()
+		pollInterval       = app.Flag("poll", "Poll interval controls how often an individual resource should be checked for drift.").Default("10m").Duration()
```
ulucinar, line 41: Why are we increasing the poll interval from 1m to 10m?

cmd/provider/main.go
```diff
@@ -91,8 +93,8 @@ func main() {
 		Options: xpcontroller.Options{
 			Logger:                  log,
 			GlobalRateLimiter:       ratelimiter.NewGlobal(*maxReconcileRate),
-			PollInterval:            1 * time.Minute,
-			MaxConcurrentReconciles: 1,
+			PollInterval:            *pollInterval,
+			MaxConcurrentReconciles: *maxReconcileRate,
```
ulucinar, line 97: Should we have separate parameters for the max reconcile rate (which is a global rate shared between all reconcilers) and the max concurrent reconciles (which is a count of workers consuming the workqueue of a controller)?

turkenh: @ulucinar I would like to be consistent with other providers and @negz has a detailed description [here](https://github.com/crossplane/crossplane/issues/2595) on why it makes sense to use the same parameter for both.

cmd/provider/main.go
turkenh: This is the proposed value after my experiments in https://github.com/upbound/upjet/issues/116#issuecomment-1308941093 with the motivation below:

In upjet based controllers a typical reconcile loop tooks longer than a native provider due to underlying implementation. If we go with a polling interval like `1m`, we immediately cause the work queue to fill up especially when there are multiple instances of the same kind. I believe it is more reasonable to default to a higher value if a typical reconcile loop takes is in the order of ~10s (i.e. refresh/plan).

cmd/provider/main.go
ulucinar: Thanks for the pointer @turkenh. I agree, we had better follow the convention suggested there.

ulucinar (approved):
Thanks @turkenh, lgtm.

PR #176, opened 2022-11-23
Add a "Show diff" step to show diffs in case the make check-diff step fails

ulucinar:

This PR adds a `Show diff` step that's run if the `check-diff` step fails in the `CI` pipeline to show what content changes are failing the job.

Also, while working in the context of updating the Terraform provider plugin version, I realized we do not honor the declared Terraform provider version if the working directory for the Terraform provider repo already exists (under `.work`). Previously we were [removing](https://github.com/upbound/official-providers/blob/2fc9ad914fb2132dfdc9b4a2c4c91f90000cc5bd/scripts/scrape_metadata.sh#L8) this work directory as we are doing a sparse checkout. So, applied the same approach here.

I have:

Please see here:
https://github.com/upbound/provider-aws/actions/runs/3533092551/jobs/5928276392

And in case `check-diff` does not fail, this step is skipped:
https://github.com/upbound/provider-aws/actions/runs/3533118984/jobs/5928338725

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,6 +118,7 @@ jobs:
         run: make vendor vendor.check
 
       - name: Check Diff
+        id: check-diff
         run: |
           mkdir _output
           make check-diff
@@ -125,6 +126,10 @@ jobs:
           # check-diff depends on the generate Make target, and we would like
           # to save a skipped resource list
           SKIPPED_RESOURCES_CSV: ../_output/skipped_resources.csv
+
+      - name: Show diff
+        if: failure() && steps.check-diff.outcome == 'failure'
+        run: git diff
 
       - name: Publish skipped resources CSV to Github
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -126,6 +126,10 @@ jobs:
           # to save a skipped resource list
           SKIPPED_RESOURCES_CSV: ../_output/skipped_resources.csv
 
+      - name: Show diff
+        if: failure() && steps.check-diff.outcome == 'failure'
+        run: git diff
```
turkenh, line 131: Hmm, wouldn't this help already? https://github.com/upbound/build/pull/168

ulucinar: Thank you for the pointer @turkenh. I checked what's implemented there. Looks like we would still be missing the file content diff output I want to achieve with that change.

A couple of times, I've run into situations where I did not know what the content diff was as a result of `make generate` run. We have also discussed offline whether it would make sense to make this functionality part of the build submodule (i.e., if a `git diff` detects changes in the files, instead of reporting a status output, we could report a content output). The concern was to keep the content diff report separate in a different job step then the `Check Diff` step.

turkenh (approved):
Thanks @ulucinar!

PR #204, opened 2022-11-25
Add additional connection details for MQ (RabbitMQ)

loafoe:

Add additional connection details for MQ (RabbitMQ) instances:

- instance_0_endpoint_0
- instance_0_console_url

I have:

Test by building and deploying to an AWS EKS cluster and provisioning an AmazonMQ RabbitMQ cluster and retrieving the new connection details for use in an app

### DIFF
--- a/config/mq/config.go
+++ b/config/mq/config.go
@@ -5,6 +5,8 @@ Copyright 2022 Upbound Inc.
 package mq
 
 import (
+	"fmt"
+
 	"github.com/upbound/upjet/pkg/config"
 )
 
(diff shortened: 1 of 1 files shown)


config/mq/config.go
```diff
@@ -17,5 +17,20 @@ func Configure(p *config.Provider) {
 		//  float64 field. Thus here we remove the automatically injected
 		//  cross-resource reference from example manifests.
 		delete(r.References, "configuration.revision")
+
+		r.Sensitive.AdditionalConnectionDetailsFn = func(attr map[string]any) (map[string][]byte, error) {
+			conn := map[string][]byte{}
+			if instances, ok := attr["instances"].([]any); ok && len(instances) > 0 {
+				if instance, ok := instances[0].(map[string]any); ok {
```
ulucinar, line 24: Hi @loafoe,
Thank you for the PR. Are we not interested in exposing the details of the other instances if `len(instances) > 1` or can there only be at most one instance available here?

loafoe: We have not seen provisions with more than one instance. With that said I was thinking of post fixing the key with the index, just to be future proof. Any suggestion on this?

ulucinar: Hi @loafoe,
Yes, agreed, that would be great. We can iterate over the `instances` and export each instance's `console_url` and `enpoint` by post fixing them with the `instances` array index.

ulucinar: The connection details secret is part of the MR's public interface so either we can now just use `console_url_0` and `endpoint _0` as connection details secret keys but we do not expose the other items in the `instances` array, or we can implement the for loop and expose all the items in `instances` using secret data keys with the indices.

sergenyalcin: @loafoe How do you feel about @ulucinar's suggestion?

loafoe: @ulucinar @sergenyalcin modified the mapping:

For RabbitMQ:

`instance_0_console_url`
`instance_0_endpoint_0`

For ActiveMQ:

`instance_0_console_url`
`instance_0_endpoint_0`
`instance_0_endpoint_1`
...
`instance_0_endpoint_4`

This captures all data from the Terraform state and allows usage of the broker without a composition.

ulucinar (approved):
Thanks @loafoe, lgtm.

PR #282, opened 2022-12-16
Configure some resources without testing

sergenyalcin:

Configure some resources without testing

I have:


ulucinar (approved):
Thanks @sergenyalcin. Unfortunately without doing the actual tests and observing IDs, we cannot always normalize the APIs. There are also certain cases in which we can have further normalization in the external-name itself (and in certain cases the [config.TemplatedStringAsIdentifierWithNoName](https://github.com/upbound/provider-aws/blob/d455989b0942116e7a34276bd37cf364aa985e0e/config/externalname.go#L1796) helps but sometimes Terraform validation kicks in and we cannot use it (we need to check the implementation for further improvement). These currently look like to be safe choices but we should be keeping an eye while these resources are being tested for API normalization opportunities (and secondarily external-name normalizations).

PR #284, opened 2022-12-16
Configure some resources without testing II

sergenyalcin:

Configure some resources without testing part II

I have:

### DIFF
--- a/config/externalnamenottested.go
+++ b/config/externalnamenottested.go
@@ -528,4 +528,101 @@ var ExternalNameNotTestedConfigs = map[string]config.ExternalName{
 	// No import
 	// TODO: For now API is not normalized. While testing resource we can check the actual ID and normalize the API.
 	"aws_macie_s3_bucket_association": config.IdentifierFromProvider,
+
+	// macie2
+	//
+	// aws_macie2_account can be imported using the id
+	"aws_macie2_account": config.IdentifierFromProvider,
+	// aws_macie2_classification_job can be imported using the id
+	"aws_macie2_classification_job": config.IdentifierFromProvider,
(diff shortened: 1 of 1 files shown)


config/externalnamenottested.go
```diff
@@ -394,4 +394,101 @@ var ExternalNameNotTestedConfigs = map[string]config.ExternalName{
...
+	// Use the user_name to import a user
+	"aws_memorydb_user": config.ParameterAsIdentifier("user_name"),
+
+	// opsworks
+	//
+	// Opsworks Application can be imported using the id
+	"aws_opsworks_application": config.IdentifierFromProvider,
+	// OpsWorks Custom Layers can be imported using the id
+	"aws_opsworks_custom_layer": config.IdentifierFromProvider,
+	// No import
```
ulucinar, line 453: Thanks @sergenyalcin for taking note of these TODOs. Let's do our best to enforce these during testing & reviews.

ulucinar (approved):
Thank you @sergenyalcin. Leaving the same comment for this PR also:
Unfortunately without doing the actual tests and observing IDs, we cannot always normalize the APIs (lack of documentation). There are also certain cases in which we can have further normalization in the external-name itself (and in certain cases the [config.TemplatedStringAsIdentifierWithNoName](https://github.com/upbound/provider-aws/blob/d455989b0942116e7a34276bd37cf364aa985e0e/config/externalname.go#L1796) helps but sometimes Terraform validation kicks in and we cannot use it (we need to check the implementation for further improvement). These currently look like to be safe choices but we should be keeping an eye while these resources are being tested for API normalization opportunities (and secondarily external-name normalizations).

PR #445, opened 2023-01-13
Make endpoint available for RDS DB Instance

### LINKED ISSUE #442: rds: WriteConnectionSecretToRef not working
### What happened?

I am creating a Postgres instance with the rds api. I have set:
`  writeConnectionSecretToRef:
    name: db-conn
    namespace: my-namespace`

For some reason the only data being written to the secret is `attribute.password`. The Postgres instance is creating successfully and has an endpoint.

I need the endpoint of the Postgres instance in the secret. This was working on the `database.aws.crossplane.io/v1beta1` api version.

I also tried using the `publishConnectionDetailsTo` spec field which failed to create any secret at all.

### How can we reproduce it?

If I delete the Postgres instance and recreate it the secret will be created, but only with the password data.

* Universal Crossplane Version: 1.10.1
* Provider Version: 0.26.0

loafoe:

Fixes #442 

I have:

Provisioned  `Instance` using crossplane `1.10.1` and a local build, verified `endpoint` value was available

### DIFF
--- a/config/rds/config.go
+++ b/config/rds/config.go
@@ -74,6 +74,13 @@ func Configure(p *config.Provider) {
 		r.LateInitializer = config.LateInitializer{
 			IgnoredFields: []string{"name", "db_name"},
 		}
+		r.Sensitive.AdditionalConnectionDetailsFn = func(attr map[string]any) (map[string][]byte, error) {
+			conn := map[string][]byte{}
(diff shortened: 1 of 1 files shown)


ytsarev (approved):
lgtm, thanks for the fix!

sergenyalcin (approved):

PR #451, opened 2023-01-17
allow using assumeRoleChain with IAM resources

### LINKED ISSUE #215: iam: Role CRD is not working with ProviderConfig with assumeRoleChain
### What happened?

Creation of IAM Roles failed. It seems to be due to operation error STS (I use an IAM user assuming a role in other AWS account). The error message is:
"cannot get terraform setup: failed to retrieve aws credentials from aws config: failed to refresh cached credentials, operation error STS: AssumeRole, failed to resolve service endpoint, an AWS region is required, but was not found"

The others ressources in my composition are well created (VPC, Subnet, NATGateway, InternetGateway, etc...)

### How can we reproduce it?

ProviderConfig:
```yaml
kind: ProviderConfig
metadata:
  name: test1
spec:
  assumeRoleChain:
    - roleARN: arn:aws:iam::22222222222:role/admin
  credentials:
    secretRef:
      key: creds
      name: aws-creds
      namespace: test1
    source: Secret
```

Role:
```yaml
 name: iamrole-controlplane
    base:
      apiVersion: iam.aws.upbound.io/v1beta1
      kind: Role
      spec:
        forProvider:
          assumeRolePolicy: |
            {
              "Version": "2012-10-17",
              "Statement": [
                  {
                      "Effect": "Allow",
                      "Principal": {
                          "Service": [
                              "eks.amazonaws.com"
                          ]
                      },
                      "Action": [
                          "sts:AssumeRole"
                      ]
                  }
              ]
            }
    patches:
    - type: PatchSet
      patchSetName: providerconfigref
    - fromFieldPath: spec.id
      toFieldPath: metadata.name
      transforms:
      - type: string
        string:
          fmt: '%s-controlplane'
    - fromFieldPath: spec.id
      toFieldPath: metadata.labels.role
      transforms:
      - type: string
        string:
          fmt: '%s-controlplane'
```

* Universal Crossplane Version: 1.10.1-up.1
* Provider Version: xpkg.upbound.io/upbound/provider-aws:v0.21.0
* kubectl version: 1.22.5
* Kubernetes version: docker-desktop Kubernetes v1.22.5
* OS: macOS 12.6 Monterey

ctkeyser:

Fixes #215

currently assumeRoleChain doesn't work for IAM resources because IAM resources don't require a region. the error is
```
cannot get terraform setup: failed to retrieve aws credentials from aws config: failed to refresh cached credentials, operation error STS: AssumeRole, failed to resolve service endpoint, an AWS region is required, but was not found
```
this change adds a fallback to "aws-global" region for STS requests in cases where no region is set in the managed resource

I have:

I built the provider w/ the fix, installed it in my environment, and confirmed I no longer get the 'region is required' error

### DIFF
--- a/internal/clients/provider_config.go
+++ b/internal/clients/provider_config.go
@@ -264,9 +264,14 @@ func UseProviderSecret(ctx context.Context, data []byte, profile, region string)
 // AssumeRoleWithWebIdentity & AssumeRoles.
 func GetRoleChainConfig(ctx context.Context, pcs *v1beta1.ProviderConfigSpec, cfg *aws.Config) (*aws.Config, error) {
 	pCfg := cfg
+	regionOpt := func(o *sts.Options) {
+		if cfg.Region == "" {
+			o.Region = GlobalRegion
+		}
+	}
 	for _, aro := range pcs.AssumeRoleChain {
 		stsAssume := stscreds.NewAssumeRoleProvider(
-			sts.NewFromConfig(*pCfg), //nolint:contextcheck
+			sts.NewFromConfig(*pCfg, regionOpt), //nolint:contextcheck
 			aws.ToString(aro.RoleARN),
 			SetAssumeRoleOptions(aro),
 		)


apis/v1beta1/types.go
```diff
@@ -17,6 +17,10 @@ type ProviderConfigSpec struct {
 	// AssumeRoleChain defines the options for assuming an IAM role
 	AssumeRoleChain []AssumeRoleOptions `json:"assumeRoleChain,omitempty"`
 
+	// AssumeRoleSTSRegion defines the region used for STS requests. If not defined,
+	// the managed resource region will be used if available
+	AssumeRoleSTSRegion *string `json:"assumeRoleSTSRegion,omitempty"`
```
ulucinar, line 22: An alternative to consider would be to add the the STS region option to the `v1beta1.AssumeRoleOptions` type. Then, we the `ProviderConfig` would look like:
```yaml
...
spec:
  assumeRoleChain:
  - stsRegion: us-west-1
    roleARN: <role ARN 1>
  - stsRegion: us-west-1
    roleARN: <role ARN 2>
  credentials:
    secretRef:
      key: creds
      name: aws-creds
      namespace: upbound-system
    source: Secret
```

However, you'd need to configure an STS region for each role in the chain if you are provisioning a resource without a region (e.g., IAM role) and not sure if there's utility in configuring different STS regions for different roles in the chain. Of course, if needed, we could have both levels of API exposed, e.g., a `spec.assumeRoleSTSRegion` and `spec.assumeRoleChain[].stsRegion` (which we could add later if needed). So, I think the proposed API is fine. What do you think?

ulucinar, line 22: nit: We may give an example of the new API under `examples/providerconfig/`, e.g., in `examples/providerconfig/user-creds-with-assume-role.yaml` or in a new example file.

ulucinar:
Hi @ctkeyser,
Thank you for the proposal. I think it looks fine but left some review comments so that we discuss some alternatives. Also left a comment [here](https://github.com/upbound/provider-aws/issues/215#issuecomment-1401651493) regarding the underlying issue.

ulucinar:

apis/v1beta1/types.go
ctkeyser: my first idea was also to put it in `v1beta1.AssumeRoleOptions`, but then I moved it because I didn't think there would be a need to chain assume role requests across different regions. I definitely would have rather put it inside an assumeRole config, but the current structure doesn't really support that. so I think this is probably the best option

ctkeyser: yeah definitely

ctkeyser:
thank you for the review @ulucinar. let me know if you have more thoughts on this

ulucinar:

ulucinar (approved):
Thank you @ctkeyser for fixing the issue with provisioning IAM resources when using STS role chaining, lgtm.

apis/v1beta1/types.go
ulucinar: As a side note for documentation purposes, we had also previously discussed defaulting to a region specified in the `ProviderConfig` if a managed resource does not specify an optional `spec.forProvider.region`. The discussion's context was not this PR's target issue (but rather the UX around provider-aws MR APIs). But looks like it's also related to the current discussion.

internal/clients/provider_config.go
```diff
@@ -264,9 +264,14 @@ func UseProviderSecret(ctx context.Context, data []byte, profile, region string)
 // AssumeRoleWithWebIdentity & AssumeRoles.
 func GetRoleChainConfig(ctx context.Context, pcs *v1beta1.ProviderConfigSpec, cfg *aws.Config) (*aws.Config, error) {
 	pCfg := cfg
+	regionOpt := func(o *sts.Options) {
+		if pcs.AssumeRoleSTSRegion != nil {
+			o.Region = *pcs.AssumeRoleSTSRegion
+		}
+	}
```
hasheddan, line 271: @ctkeyser @ulucinar I wonder if it would make sense to fall back to the AWS global partition here if an STS region is not defined? The AWS SDK itself [will do so](https://github.com/aws/aws-sdk-go-v2/issues/1778) if it does not recognize the region that is passed, but will error if no region is passed. Thoughts?

ulucinar: Hi @hasheddan,
Makes sense to me.

ctkeyser: that sounds good to me, but I'm not sure what that change looks like. I tried changing region to `global` and `aws-global` and I ended up getting errors saying dns doesn't resolve for `sts.aws-global.amazonaws.com` or `sts.global.amazonaws.com`. dns does resolve for `sts.us-west-1.amazonaws.com` though

I also don't see STS in this list of global endpoints https://docs.aws.amazon.com/general/latest/gr/rande.html#global-endpoints

hasheddan: @ctkeyser I believe it may just be `sts.amazonaws.com` https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html

ctkeyser: ok I tested just falling back to `aws-global` if no region set, and that does work in my environment:
```diff
--- a/internal/clients/provider_config.go
+++ b/internal/clients/provider_config.go
@@ -265,8 +265,8 @@ func UseProviderSecret(ctx context.Context, data []byte, profile, region string)
 func GetRoleChainConfig(ctx context.Context, pcs *v1beta1.ProviderConfigSpec, cfg *aws.Config) (*aws.Config, error) {
        pCfg := cfg
        regionOpt := func(o *sts.Options) {
-               if pcs.AssumeRoleSTSRegion != nil {
-                       o.Region = *pcs.AssumeRoleSTSRegion
+               if cfg.Region == "" {
+                       o.Region = GlobalRegion
                }
        }
        for _, aro := range pcs.AssumeRoleChain {
```
I was able to assume a role and create an IAM resource with that
I also tried changing getRegion to return GlobalRegion if no region was found in the managed resource, but that didn't work. got the error:
```
managed/iam.aws.upbound.io/v1beta1, kind=role  cannot run refresh: refresh failed: Invalid AWS Region: aws-global
```

do you guys think this is good enough and we don't need the providerconfig option? so basically any region-less managed resource will use aws-global for STS. that seems simple enough to me

hasheddan: This seems reasonable to me -- we can always expand the API of `ProviderConfig` if needed. So in summary:
- If an assume role chain is being used and the MR specifies a region, we will use that region endpoint for the STS operations.
- If chain is being used and the MR does not specify a region, we use the global STS endpoint.

I think potentially using a separate region's STS endpoint could be desired by a user, but let's wait until it is requested since it involves API expansion. Are you aligned @ulucinar?

ulucinar: I don't much about the use cases here. But one hypothetical scenario I may imagine is what happens if networking access to the host `sts.amazonaws.com` is blocked (due to law, temporary issue, etc.). But my understanding is that it (role chaining) was not working up to now for region-less resources and if we make a backward-compatible change (i.e., only revert back to `sts.amazonaws.com` for role chaining if the MR does not specify a region), we should not be breaking any existing (already working) scenarios and would be solving the problem for at least a subset of the users (of provider-aws). 

So, I think we are good to go. But to be on the safe side, even if, currently, we will not allow configuration of the STS region for region-less resources, let's scope the change only for region-less resources (so that the change is backward-compatible as discussed above).

ctkeyser: ok I pushed the change. it only sets the region to aws-global for STS requests when no region is set, so it should only occur in situations where it currently doesn't work at all

ulucinar:

internal/clients/provider_config.go
```diff
@@ -264,9 +264,14 @@ func UseProviderSecret(ctx context.Context, data []byte, profile, region string)
 // AssumeRoleWithWebIdentity & AssumeRoles.
 func GetRoleChainConfig(ctx context.Context, pcs *v1beta1.ProviderConfigSpec, cfg *aws.Config) (*aws.Config, error) {
 	pCfg := cfg
+	regionOpt := func(o *sts.Options) {
+		if cfg.Region == "" {
+			o.Region = GlobalRegion
+		}
+	}
```
hasheddan, line 271: might be nice to pull this out into something like:
```go
NewRegionFallbackOpt(region string) func(*sts.Options) {
  if region != "" {
    return func(*sts.Options) {}
  }
  return func(o *sts.Options) {
    o.Region = GlobalRegion
  }
}
```

But I am fine with deferring doing so for now.

hasheddan (approved):
One comment but non-blocking -- thanks @ctkeyser! Might also be worth squashing commit history here 👍🏻

ctkeyser:
@hasheddan should I squash and force-push or is squash+merge preferred?

hasheddan:
@ctkeyser squash and force push would be great if you don't mind!

hasheddan:
I'll go ahead and merge here since @ulucinar previously approved and indicated alignment on the updated direction 👍🏻

PR #476, opened 2023-01-31
Enable `wafv2_web_acl_association`

smcavallo:
Signed-off-by: smcavallo <smcavallo@hotmail.com>

Currently `upbound/provider-aws` can create an ALB and create a WAF but it cannot associate the WAF with any resources.

Adds support for `wafv2_web_acl_association` and associating a WAF to an ALB:
See - https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_web_acl_association

Note that the property named `resource_arn` is generic and could potentially resolve to an ARN of an Application Load Balancer, an Amazon API Gateway stage, or an Amazon Cognito User Pool.
Currently in this PR only resolving to an ALB is supported which is most likely the most common use case.
 I thought there might be a helper for resolving to different resources depending on the ARN provided in the common helper class https://github.com/upbound/provider-aws/blob/main/config/common/common.go but it doesn't look like this is supported and there are no other `resource_arn` resolvers which look like they currently support this.  (Maybe this is configured via `FormattedIdentifierFromProvider` or `TemplatedStringAsIdentifier` ?)

Fixes #

I have:

- [X ] Run `make reviewable test` to ensure this PR is ready for review.

Locally e2e


sergenyalcin:
Thanks @smcavallo for this PR. We want to run automated tests for this PR. Could you please copy the manifest in the `examples-generated` to the `examples` folder? Then I can run the automated tests and then let's see what is happening.

smcavallo:
thanks @sergenyalcin - example has been added for automated tests

turkenf:
Require status checks changed could you please rebase this PR?

config/wafv2/config.go
```diff
@@ -0,0 +1,18 @@
+package wafv2
+
+import "github.com/upbound/upjet/pkg/config"
+
+// Configure adds configurations for wafv2 group.
+func Configure(p *config.Provider) {
+	p.AddResourceConfigurator("aws_wafv2_web_acl_association", func(r *config.Resource) {
+		r.References = config.References{
+			"resource_arn": config.Reference{
```
ulucinar, line 9: [Looks like](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_web_acl_association#resource_arn) `resource_arn` may refer to ARNs of multiple kinds of resources (Application Load Balancer, an Amazon API Gateway stage, or an Amazon Cognito User Pool). So, we should not configure a referencer here.

examples/wafv2/webaclassociation.yaml
```diff
@@ -0,0 +1,95 @@
...
+  labels:
+    testing.upbound.io/example-name: example
+  name: example
+spec:
+  forProvider:
+    region: us-west-1
+    resourceArnSelector:
+      matchLabels:
+        testing.upbound.io/example-name: example
+    webAclArnSelector:
```
ulucinar, line 15: Unless we also configure `wafv2.WebACL` and reference it here, this example manifest will not be testable on its own. If you decide not to configure `wafv2.WebACL` within this PR, then we will ask you to manually test `wafv2. WebACLAssociation` and report the results in the PR description and add the `upjet.upbound.io/manual-intervention` [with its reasoning](https://github.com/upbound/provider-aws/blob/ecb2053d23a951507c8ea03ea195d7d9db5c5420/examples/ec2/tag.yaml#L5). Or, if you decide to also configure the `wafv2.WebACL`, then we can add it as a dependency in this manifest and [uptest this resource](https://github.com/upbound/uptest). We will help you in uptesting this resource during the reviews if you choose to follow that path.

config/wafv2/config.go
turkenf: @ulucinar same question here: https://github.com/upbound/provider-aws/pull/478#discussion_r1223018144, can we still configure the reference resolver to the most common resource?

PR #477, opened 2023-02-01
Add route53_zone_association resource

sergenyalcin:

Fixes #https://github.com/upbound/provider-aws/issues/393

I have:


sergenyalcin:
/test-examples="examples/route53/zoneassociation.yaml"

config/route53/config.go
```diff
@@ -53,8 +53,6 @@ func Configure(p *config.Provider) {
 		}
 	})
 	p.AddResourceConfigurator("aws_route53_zone", func(r *config.Resource) {
-		// Mutually exclusive with aws_route53_zone_association
```
ulucinar, line 56: Let's add a comment why we are violating the XRM here: If we don't allow a VPC to be specified here, then, to the best of our knowledge, we are losing functionality to provision private zones and you can only associate a VPC with a zone if the zone is private. And to the best of our knowledge, Zone's `vpc` API is the only one that allows us to make a Zone private.

ulucinar:
I think we are tackling the same problem [here](https://github.com/upbound/provider-aws/pull/456). Let's review that PR.

PR #530, opened 2023-02-16
add more fields to connection details

alexinthesky:

adds more fields  ( host , address, port )for rds instance connection details. We need this to pass them along to provider-sql and the upcoming upjet [generated provider-postgresql](https://github.com/alexinthesky/provider-postgresql) ). I guess this comes handy as the tf providerconfig asks for separate host and port, as most of the PG clients )

Fixes #

I have:

ran it locally against

### DIFF
--- a/config/rds/config.go
+++ b/config/rds/config.go
@@ -94,8 +94,9 @@ func Configure(p *config.Provider) {
 			if a, ok := attr["username"].(string); ok {
 				conn["username"] = []byte(a)
 			}
-
-			conn["port"] = []byte(fmt.Sprintf("%v", attr["port"]))
+			if a, ok := attr["port"]; ok {
+				conn["port"] = []byte(fmt.Sprintf("%v", a))
+			}
 
 			return conn, nil
 		}
--- a/config/rds/config.go
+++ b/config/rds/config.go
@@ -8,6 +8,8 @@ import (
 	"github.com/upbound/upjet/pkg/config"
 
 	"github.com/upbound/provider-aws/config/common"
+
+	"fmt"
 )
 
 // Configure adds configurations for rds group.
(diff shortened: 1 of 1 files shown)


jeanduplessis:
@sergenyalcin would you please do a review here.

alexinthesky:
@sergenyalcin @ulucinar any chance you guys can look into this please?

config/rds/config.go
```diff
@@ -85,9 +87,16 @@ func Configure(p *config.Provider) {
...
 			}
+			if a, ok := attr["address"].(string); ok {
+				conn["address"] = []byte(a)
+				conn["host"] = []byte(a)
+			}
 			if a, ok := attr["username"].(string); ok {
 				conn["username"] = []byte(a)
 			}
+
+			conn["port"] = []byte(fmt.Sprintf("%v", attr["port"]))
```
sergenyalcin, line 98: As far as I understand, we assume that `port` field of attr map will be full. How about handling a case where this value is null? Like that:

```go
if a, ok := attr["port"]; ok {
				conn["port"] = []byte(fmt.Sprintf("%v", a))
			}
```

sergenyalcin (commented):
Thanks @alexinthesky I left a comment

config/rds/config.go
alexinthesky: 👍

alexinthesky: comment taken

sergenyalcin:
`CI / lint` job fails with a timeout after a few retries. I ran the lint in my local and it passed successfully. I will re-trigger this job for a few times

sergenyalcin (approved):
Thanks @alexinthesky LGTM!

PR #593, opened 2023-03-07
Default STS region to global for WebIdentity

hasheddan:

Updates the STS client region for WebIdentity role assumption to fall back to global if no region is provided for the managed resource being reconciled. This matches the existing behavior in the STS client construction for assuming a role via chaining.

Follow up to https://github.com/upbound/provider-aws/pull/451

I have:

Verified that the same error experienced in #451 is no longer showing up on IAM Role resources when using the following `ProviderConfig` (which utilizes web identity):

```
kind: ProviderConfig
metadata:
  name: default
spec:
  credentials:
    source: Upbound
    upbound:
      webIdentity:
        roleARN: arn:aws:iam::REDACTED:role/REDACTED
```


ulucinar (approved):
Thanks @hasheddan, lgtm.

PR #595, opened 2023-03-08
S3: BucketACL: Enable canned ACL support

### LINKED ISSUE #561: Update BucketACL to allow canned acls
I am migrating some resources from community provider-aws and some from terraform to upbound provider-aws.  Both [community provider-aws](https://doc.crds.dev/github.com/crossplane/provider-aws/s3.aws.crossplane.io/Bucket/v1beta1@v0.37.1#spec-forProvider-acl) and [terraform](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_acl) provide easy way to set canned acl.  When managing s3 bucket for both community provider and terraform I can use `acl: private` to set the acl for a bucket.  

For upbound aws provider I have to get `aws_canonical_user_id.current.id` as an extra step but there is no way to get it natively and I would have to use external script to get the id and pass them in to [BucketACL](https://marketplace.upbound.io/providers/upbound/provider-aws/v0.30.0/resources/s3.aws.upbound.io/BucketACL/v1beta1)

There is more detail in [this slack thread](https://crossplane.slack.com/archives/C01TRKD4623/p1677101356835579) as well.

For BucketACL we need `forProvider.acl` that accepts canned acls.

For BucketACL we need `forProvider.acl` that accepts canned acls.

ytsarev:

Add [Canned ACL](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html#canned-acl) support to BucketACL resource

Required for manageable ACL control of S3 buckets.

Fixes #561

I have:

* Apply new canned ACL example
```
k apply -f examples/s3/bucketaclcanned.yaml
NAME                                  READY   SYNCED   EXTERNAL-NAME                AGE
bucketacl.s3.aws.upbound.io/example   True    True     testtestest-canned,private   77s

NAME                                          READY   SYNCED   EXTERNAL-NAME        AGE
bucket.s3.aws.upbound.io/testtestest-canned   True    True     testtestest-canned   77s
```

* Regression test `accessControlPolicy`
```
apiVersion: s3.aws.upbound.io/v1beta1
kind: BucketACL
metadata:
  name: example
  annotations:
    upjet.upbound.io/manual-intervention: "Requires adding canonical ID of the owner user who created the Bucket."
spec:
  forProvider:
    region: us-west-1
    bucketSelector:
      matchLabels:
        testing.upbound.io/example-name: s3
    accessControlPolicy:
      -  grant:
           -  permission: "READ"
              grantee:
                - id: <redacted>
                  type: CanonicalUser
         owner:
           -  id: <redacted>
```
```
k get managed
NAME                                           READY   SYNCED   EXTERNAL-NAME         AGE
bucket.s3.aws.upbound.io/testcannednotcanned   True    True     testcannednotcanned   6m55s

NAME                                  READY   SYNCED   EXTERNAL-NAME         AGE
bucketacl.s3.aws.upbound.io/example   True    True     testcannednotcanned   48s
```


ytsarev:
/test-examples="examples/s3/bucketaclcanned.yaml"

examples/s3/bucketaclcanned.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: s3.aws.upbound.io/v1beta1
```
ulucinar, line 1: When we start triggering automated tests with uptest (e.g., by testing the `examples/s3/bucketacl.yaml` file), this example will not be covered. We can also reconsider how to trigger the tests automatically but it simplifies to rely on the examples/\<group\>/\<resource\>.yaml convention. 

We don't need to decide on how to improve this in this PR, but how about moving this example into the `examples/s3/bucketacl.yaml` file if they can share common dependencies? Or maybe better, we can have another convention that will help us detecting what to test. A proposal would be to name the file with something like `bucketacl-canned.yaml`. The reasoning is `bucketaclcanned` could be another resource's name but we can easily deduce the test target resource name with this convention. We could as well just the file name as `bucketaclcanned.yaml` but in our tooling inspect the resource manifests in the file contents and deduce the test target resource (we were using a specific test annotation to mark the test target previously, or we can just assume the first manifest is the test target manifest). This way we would not need any conventions for the file names. I lean towards a convention like `bucketacl-canned.yaml` as it looks simpler to me. What do you think? cc. @sergenyalcin

config/s3/config.go
```diff
@@ -32,7 +32,9 @@ func Configure(p *config.Provider) {
 	})
 
 	p.AddResourceConfigurator("aws_s3_bucket_acl", func(r *config.Resource) {
-		config.MoveToStatus(r.TerraformResource, "acl")
```
ulucinar, line 35: Hi @ytsarev,
My understanding is that this is another case where we previously removed a field from the API because there's another one that we though does the same job (e.g., in this case I believe it's the `spec.forProvider. accessControlPolicy`) but it turns out that the remaining API is not sufficient to handle certain use cases (in this specific case, being able to define canned ACLs). Is this correct?

config/s3/config.go
```diff
@@ -32,7 +32,9 @@ func Configure(p *config.Provider) {
...
+		r.LateInitializer = config.LateInitializer{
+			IgnoredFields: []string{"acl", "access_control_policy"},
```
ulucinar, line 36: Taking a note also in this PR to advocate for this. We can consider handling these conflicts automatically in upjet as the schema already has this information:
https://github.com/hashicorp/terraform-provider-aws/blob/baf25d816f9f2d5c463ab0d9bd267d28d15d1674/internal/service/s3/bucket_acl.go#L41
Will open a separate issue in upjet.

ulucinar (approved):
Thank you @ytsarev, lgtm. We may just consider adding the example-id annotations in the new examples, and left some other comments for discussion and clarification.

config/s3/config.go
ytsarev: @ulucinar that's correct,[canned ACLs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html#canned-acl) are simple strings, aka prepackaged ACLs by the cloud provider. `accessControlPolicy` is a full-fledged ACL doc for the granular control. We should not force users to use a more complex `accessControlPolicy` when it is actually not required

config/s3/config.go
ytsarev: @ulucinar that will be amazing, Alper, thanks!

examples/s3/bucketaclcanned.yaml
ytsarev: @ulucinar that makes sense, I can rename the file. Do we actually have a code for mass example execution with uptest? Currently I am aware of targeted per example one only

examples/s3/bucketaclcanned.yaml
```diff
@@ -0,0 +1,23 @@
...
+kind: BucketACL
+metadata:
+  name: example
```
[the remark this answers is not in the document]
ytsarev: will do

ytsarev:
/test-examples="examples/s3/bucketacl-canned.yaml"

examples/s3/bucketaclcanned.yaml
ulucinar: Yes, group-wide test executions (like specifying `examples/ec2` as the test target) have proven to be problematic. So, we are triggering fine-granular, specific resource-targeted tests (like `examples/s3/bucketacl.yaml`) as you mentioned. Even if we adopt the naming convention suggested above, we would still need changes in our tooling (either in upjet itself, or in the clients of upjet) so that when we intent to test the `s3.BucketACL` managed resource implementation, we would use manifests from `bucketacl.yaml` and `bucketacl-*.yaml` (this is not supported yet). 

The idea is to have all the use cases captured in the example manifests to be covered by our automated tests. The way to do this is right now putting all examples into a single file (e.g., `examples/s3/bucketacl.yaml`), excluding the directory-targeted tests. But this does not scale well in my opinion, and having per use-case example manifests looks better to me. This is something we need to consider when we continue with the triggering tests automatically work stream (epic).

ytsarev:
/test-examples="examples/s3/bucketaclcanned.yaml"

ytsarev:
/test-examples="examples/s3/bucketacl-canned.yaml"

PR #609, opened 2023-03-17
Remove workflow update

dverveiko:
As we change or CI/CD process to use reusable workflows, we need to remove updating of old workflow from renovate-bot
configuration

I have:


ulucinar (approved):
Thanks @dverveiko, lgtm. We have a GO_VERSION declaration in uptest-all.yml workflow file but most probably we will remove this. And if, for some reason, decide to keep it, then we can consider moving it to uptest. So I think we are good to go with this PR.

PR #616, opened 2023-03-20
Move authentication down to Terraform

### LINKED ISSUE #613: Move authentication down to Terraform

Currently, we are doing token exchange for IRSA & WebIdentity authentication [in the Crossplane provider](https://github.com/upbound/provider-aws/blob/main/internal/clients/provider_config.go). For instance, when IRSA authentication is configured via a `ProviderConfig`, at each reconciliation, the Crossplane provider exchanges the web identity JWT token available in the pod with temporary role credentials by calling the `sts.AssumeRoleWithWebIdentity`, even though a previous set of exchanged credentials may not have expired yet. This also interferes with the shared server runtime we would like to enable to improve the provider's efficiency. 

We can consider moving the exchanges down to the Terraform layer. The underlying Terraform provider is already capable of doing [AssumeRoleWithWebIdentity](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#assuming-an-iam-role-using-a-web-identity) and [AssumeRole](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#assume-role-configuration-reference). We need to check role chaining, e.g., supplying multiple `assume_role` configuration blocks. We also need to check if Terraform has already some credential caching mechanisms that we can make use of to improve efficiency. The `Upbound` authentication mechanism is also effectively a WebIdentity scheme which we can implement by setting the `AWS_WEB_IDENTITY_TOKEN_FILE` environment variable properly. Moving authentication down to the Terraform layer will also improve our security posture.

sergenyalcin:

Fixes #613 

I have:


ulucinar:

sergenyalcin:

internal/clients/aws.go
```diff
@@ -7,61 +7,160 @@ package clients
...
+			return terraform.Setup{}, errors.Wrap(err, "cannot get account id")
+		}
+		ps.ClientMetadata = map[string]string{
+			keyAccountId: account,
 		}
-		if cfg.Region == "" && mg.GetObjectKind().GroupVersionKind().Group == "iam.aws.upbound.io" {
-			cfg.Region = "us-east-1"
+
+		var err2 error
+		if (len(pc.Spec.AssumeRoleChain) > 1 || pc.Spec.Endpoint != nil) && (pc.Spec.Credentials.Source == authKeyIRSA || pc.Spec.Credentials.Source == authKeyWebIdentity || pc.Spec.Credentials.Source == authKeyUpbound) {
```
ulucinar, line 57: We should be invoking the default path if we have an AssumeRole chain of length greater than 1 or an endpoint configuration independent of the configured authentication scheme.

internal/clients/aws.go
```diff
@@ -7,61 +7,160 @@ package clients
...
-				keySessionToken:    creds.SessionToken,
-			},
-			// Account ID is not part of provider configuration schema, so it
-			// needs to be given separately.
-			ClientMetadata: map[string]string{
-				keyAccountId: *identity.Account,
-			},
-		}
-		return ps, err
+	if len(pc.Spec.AssumeRoleChain) != 0 {
```
ulucinar, line 115: We had better return an error if `len(pc.Spec.AssumeRoleChain) > 1` or there's an endpoint configuration in this path.

internal/clients/aws.go
```diff
@@ -7,61 +7,165 @@ package clients
...
+			ps, err2 = OptimizedTerraformSetupBuilder(ctx, c, mg, pc, ps)
+			if err2 != nil {
+				return terraform.Setup{}, errors.Wrap(err2, "cannot build terraform configuration")
+			}
+		}
+		return ps, nil
+	}
+}
+
+func OptimizedTerraformSetupBuilder(ctx context.Context, c client.Client, mg resource.Managed, pc *v1beta1.ProviderConfig, ps terraform.Setup) (terraform.Setup, error) { //nolint:gocyclo
```
ulucinar, line 72:
```suggestion
func OptimizedTerraformSetupBuilder(ctx context.Context, c client.Client, mg resource.Managed, pc *v1beta1.ProviderConfig, ps *terraform.Setup) error { //nolint:gocyclo
```

internal/clients/aws.go
```diff
@@ -7,61 +7,165 @@ package clients
...
+
+	if len(pc.Spec.AssumeRoleChain) != 0 {
+		ps.Configuration[keyAssumeRole] = pc.Spec.AssumeRoleChain[0].RoleARN
+		ps.Configuration[keyTags] = pc.Spec.AssumeRoleChain[0].Tags
+		ps.Configuration[keyTransitiveTagKeys] = pc.Spec.AssumeRoleChain[0].TransitiveTagKeys
+	}
+	return ps, nil
+}
+
+func DefaultTerraformSetupBuilder(ctx context.Context, c client.Client, mg resource.Managed, ps terraform.Setup) (terraform.Setup, error) {
```
ulucinar, line 128:
```suggestion
func DefaultTerraformSetupBuilder(ctx context.Context, c client.Client, mg resource.Managed, ps *terraform.Setup) error {
```

internal/clients/aws.go
```diff
@@ -7,61 +7,165 @@ package clients
...
-		return ps, err
+	if len(pc.Spec.AssumeRoleChain) > 1 || pc.Spec.Endpoint != nil {
+		return terraform.Setup{}, errors.New("shared scheduler cannot be used because the length of assume role chain array " +
+			"is more than 1 or endpoint configuration is not nil")
+	}
+
+	if len(pc.Spec.AssumeRoleChain) != 0 {
+		ps.Configuration[keyAssumeRole] = pc.Spec.AssumeRoleChain[0].RoleARN
+		ps.Configuration[keyTags] = pc.Spec.AssumeRoleChain[0].Tags
+		ps.Configuration[keyTransitiveTagKeys] = pc.Spec.AssumeRoleChain[0].TransitiveTagKeys
```
ulucinar, line 123: We also need to incorporate the `pc.Spec.AssumeRoleChain[0].ExternalID`:
image

internal/clients/aws.go
```diff
@@ -7,61 +7,165 @@ package clients
...
+		keyRegion: cfg.Region,
+	}
+
+	switch s := pc.Spec.Credentials.Source; s { //nolint:exhaustive
+	case authKeyWebIdentity:
+		if pc.Spec.Credentials.WebIdentity == nil {
+			return terraform.Setup{}, errors.New(`spec.credentials.webIdentity of ProviderConfig cannot be nil when the credential source is "WebIdentity"`)
+		}
+		ps.Configuration[keyAssumeRoleWithWebIdentity] = pc.Spec.Credentials.WebIdentity.RoleARN
+		ps.Configuration[keyWebIdentityTokenFile] = defaultIdentityTokenFile
```
ulucinar, line 87: We had better use the token file specified in the environment:
```suggestion
		ps.Configuration[keyWebIdentityTokenFile] = os.Getenv(envWebIdentityTokenFile)
```

sergenyalcin:

ulucinar:
Tested with the following provider configurations:
- IAM User access key
- IRSA
- IRSA with an AssumeRole chain of length 1
- IRSA with an AssumeRole chain of length 2
- WebIdentity -> Issue detected (empty string is not considered as a valid session name by Terraform)

internal/clients/aws.go
```diff
@@ -6,62 +6,184 @@ package clients
...
+	case authKeyWebIdentity:
+		if pc.Spec.Credentials.WebIdentity == nil {
+			return errors.New(`spec.credentials.webIdentity of ProviderConfig cannot be nil when the credential source is "WebIdentity"`)
+		}
+		if pc.Spec.Credentials.WebIdentity.RoleARN == nil {
+			return errors.New(`spec.credentials.webIdentity.roleARN of ProviderConfig cannot be nil when the credential source is "WebIdentity"`)
+		}
+		ps.Configuration[keyAssumeRoleWithWebIdentity] = map[string]any{
+			keyRoleArn:              pc.Spec.Credentials.WebIdentity.RoleARN,
+			keySessionName:          pc.Spec.Credentials.WebIdentity.RoleSessionName,
```
ulucinar, line 99: If the `pc.Spec.Credentials.WebIdentity.RoleSessionName` is an empty string, Terraform complains about it:
```shell
"Error: expected length of assume_role_with_web_identity.0.session_name to be in the range (2 - 64), got "
```

One solution could be converting the type of `pc.Spec.Credentials.WebIdentity.RoleSessionName` to `*string` but we need to make sure upgrades work smoothly (expecting empty strings to be converted to `nil` because of the `omitempty` JSON serialization tag). 

Another option could be not to put this key into the map if the value is an empty string.

internal/clients/aws.go
```diff
@@ -6,62 +6,184 @@ package clients
...
+	case authKeyUpbound:
+		if pc.Spec.Credentials.Upbound == nil || pc.Spec.Credentials.Upbound.WebIdentity == nil {
+			return errors.New(`spec.credentials.upbound.webIdentity of ProviderConfig cannot be nil when the credential source is "Upbound"`)
+		}
+		if pc.Spec.Credentials.Upbound.WebIdentity.RoleARN == nil {
+			return errors.New(`spec.credentials.upbound.webIdentity.roleARN of ProviderConfig cannot be nil when the credential source is "Upbound"`)
+		}
+		ps.Configuration[keyAssumeRoleWithWebIdentity] = map[string]any{
+			keyRoleArn:              pc.Spec.Credentials.Upbound.WebIdentity.RoleARN,
+			keySessionName:          pc.Spec.Credentials.Upbound.WebIdentity.RoleSessionName,
```
ulucinar, line 111: Did not try yet, but it should complain with the above error message given for `WebIdentity`.

internal/clients/aws.go
sergenyalcin: For the first iteration of these improvements, I preferred adding an empty string check. I think in the next iteration we may consider converting the field to `*string`.

sergenyalcin:

internal/clients/aws.go
ulucinar: Thanks @sergenyalcin. Successfully validated the following two additional provider configurations with the fix to this issue using the package `ulucinar/provider-aws-amd64:v0.31.0-ca2f21ca`:
- WebIdentity
- WebIdentity with an AssumeRole chain of length 1
- WebIdentity with an AssumeRole chain of length 2

ulucinar:
Because GitHub has now degraded availability for PRs and issues, leaving a review comment as a regular one:
internal/clients/aws.go:108
Let's try to preserve the old behavior and not introduce a breaking change here:
https://github.com/upbound/provider-aws/blob/c36cfcd775aadabfa65533014d94cffcceb0e88e/internal/clients/provider_config.go#L308

sergenyalcin:

ulucinar (approved):
Thank you @sergenyalcin, lgtm. Let's also make sure that we try the `Upbound` authentication scheme in the cloud.

PR #628, opened 2023-03-28
rds.instance: add ability to auto-generate password in referenced secret

muvaf:

Adds ability to auto-generate password. Users need to opt-in by setting `autoGeneratePassword` as `true` and should have a `Secret` reference. It will create if the referenced `Secret` does not exist and it will populate if it doesn't have the password for the given key.

I have:

Unit tests, local end-to-end test and Uptest.


muvaf:
/test-examples="examples/rds/instance.yaml"

muvaf:
/test-examples="examples/rds/instance.yaml"

muvaf:
`lint` passes locally but fails here due to https://github.com/upbound/provider-aws/issues/573

turkenh (approved):

PR #643, opened 2023-04-01
fix(ecs-service): #624 fix description of ecs service cluster field

### LINKED ISSUE #624: ECS Service cluster fields not working as documented
### What happened?

ecs-service-doc

spec.forProvider.**cluster**, spec.forProvider.**clusterSelector** , spec.forProvider.**clusterRef**  fields do not work as expected, documentation says they take an ARN, but only the cluster name works.

ecs-service-error

When ARN is given, error message is : "observe failed: cannot run refresh: refresh failed: reading ECS service (arn:aws:ecs:us-east-1:011882408883:cluster/sample-app-cluster/sample-app-service): InvalidParameterException: Unsupported resource type: cluster:"

Therefore **clusterSelector** and **clusterRef** fields that resolve to ARN are also broken and throw the same error as above.

Workaround at this time is to hardcode the cluster name into the spec.forProvider.cluster field

ecs-service-works

### How can we reproduce it?

Install newest Upbound AWS provider and apply manifests here:
https://github.com/martinflemingdev/upbound-ECS-service-cluster-field-bug

Upbound AWS Provider: [upbound/provider-aws@v0.31.0](https://marketplace.upbound.io/providers/upbound/provider-aws/v0.31.0)

haarchri:
fixed the description for ecs_service cluster field - we need the cluster name / short-id instead of ARN - ref & selector working as expected as mentioned by @dverveiko in #624 

Fixes #624

I have:

applied new example:

```
NAME                                                 READY   SYNCED   EXTERNAL-NAME             AGE
cluster.ecs.aws.upbound.io/example-cluster-service   True    True     example-cluster-service   4m2s

NAME                                         READY   SYNCED   EXTERNAL-NAME     AGE
service.ecs.aws.upbound.io/example-service   True    True     example-service   4m2s

NAME                                                           READY   SYNCED   EXTERNAL-NAME   AGE
taskdefinition.ecs.aws.upbound.io/example-service-definition   True    True     sampleservice   4m2s
```

### DIFF
--- a/examples/ecs/service-static-cluster.yaml
+++ b/examples/ecs/service-static-cluster.yaml
@@ -1,0 +1,53 @@
+apiVersion: ecs.aws.upbound.io/v1beta1
+kind: Service
+metadata:
+  name: example-service
+  annotations:
+    meta.upbound.io/example-id: ecs/v1beta1/service-static-cluster
+    upjet.upbound.io/manual-intervention: "Resource stuck in update loop, see https://github.com/upbound/provider-aws/issues/585"
+  labels:
+    testing.upbound.io/example-name: example-service
+spec:
+  forProvider:
+    taskDefinitionSelector:
+      matchLabels:
+          testing.upbound.io/example-name: example-service-definition
+    cluster: example-cluster-service
+    region: us-west-1
+    launchType: EC2
+    propagateTags: TASK_DEFINITION
+---
+apiVersion: ecs.aws.upbound.io/v1beta1
+kind: TaskDefinition
+metadata:
+  name: example-service-definition
+  annotations:
+    meta.upbound.io/example-id: ecs/v1beta1/service-static-cluster
+  labels:
+    testing.upbound.io/example-name: example-service-definition
+spec:
+  forProvider:
+    region: us-west-1
+    family: sampleservice
+    containerDefinitions: |-
+      [
+        {
+          "name": "first",
+          "image": "service-first",
+          "cpu": 10,
+          "memory": 512,
+          "essential":true
+        }
+      ]
+---
+apiVersion: ecs.aws.upbound.io/v1beta1
+kind: Cluster
+metadata:
+  annotations:
+    meta.upbound.io/example-id: ecs/v1beta1/service-static-cluster
+  labels:
+    testing.upbound.io/example-name: example-cluster-service
+  name: example-cluster-service
+spec:
+  forProvider:
+    region: us-west-1
--- a/apis/ecs/v1beta1/zz_service_types.go
+++ b/apis/ecs/v1beta1/zz_service_types.go
@@ -274,7 +274,7 @@ type ServiceParameters struct {
 	// +kubebuilder:validation:Optional
 	CapacityProviderStrategy []CapacityProviderStrategyParameters `json:"capacityProviderStrategy,omitempty" tf:"capacity_provider_strategy,omitempty"`
 
-	// ARN of an ECS cluster.
+	// Name of an ECS cluster.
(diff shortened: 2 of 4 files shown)


ytsarev (commented):
Is that actually correct? While selectors will work the actually field will take and ARN as an input https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ecs_service#cluster

haarchri:
@ytsarev check #624 the examples in the comments - you see that cluster is everytime the name instead of the arn

ytsarev (changes requested):
Got it, thanks for clarification. 

In this case, we need to extend uptest suite with an example symmetrical to https://github.com/upbound/provider-aws/blob/main/examples/ecs/service.yaml but with using an explicit cluster field name. This way we will deterministically prove that the Name/Non-ARN format works end-to-end.

Could you please add extend the PR with this additional example to cover this case?

haarchri:
I will add it

haarchri:
added example and tested:

```
kubectl get managed
Warning: BucketPolicy has been deprecated. Use spec.forProvider.policy in Bucket instead.
NAME                                                 READY   SYNCED   EXTERNAL-NAME             AGE
cluster.ecs.aws.upbound.io/example-cluster-service   True    True     example-cluster-service   4m2s

NAME                                         READY   SYNCED   EXTERNAL-NAME     AGE
service.ecs.aws.upbound.io/example-service   True    True     example-service   4m2s

NAME                                                           READY   SYNCED   EXTERNAL-NAME   AGE
taskdefinition.ecs.aws.upbound.io/example-service-definition   True    True     sampleservice   4m2s
```

ytsarev:
/test-examples="examples/ecs/service-static-cluster.yaml"

examples/ecs/service-static-cluster.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: ecs.aws.upbound.io/v1beta1
+kind: Service
+metadata:
+  name: example-service
+  annotations:
+    meta.upbound.io/example-id: ecs/v1beta1/service-static-cluster
+    upjet.upbound.io/manual-intervention: "Resource stuck in update loop"
```
ytsarev, line 7: The example is uptest green, should we remove `upjet.upbound.io/manual-intervention` annotation?

haarchri: its added because of update loop check this: https://github.com/upbound/provider-aws/issues/585

ytsarev: Ok, then
```suggestion
    upjet.upbound.io/manual-intervention: "Resource stuck in update loop, see https://github.com/upbound/provider-aws/issues/585"
```

But I am still confused because [uptest run](https://github.com/upbound/provider-aws/actions/runs/4588536248) is nicely green

haarchri: the resource overall is True / True all the time - but the update loop is running all the time / every loop

sergenyalcin (approved):
Thanks @haarchri LGTM! From my side it is good to go if @ytsarev also thinks on this way.

ytsarev:
@sergenyalcin all good in general. I still fail to understand why we need manual intervention annotation on something that proved to be automatically testable by uptest :) If somebody can explain me that simple thing I am totally happy with the merge :)

ytsarev (approved):
Ok, @sergenyalcin managed to fix my understanding. We are still skipping skipping Service involved with underlying https://github.com/upbound/uptest/blob/main/internal/prepare.go#L103 
and the rest of the resources in example are just getting successfully CRUDed by uptest. 
Sorry fo confusion, LGTM

PR #654, opened 2023-04-05
update the externalname configuration for aws_lambda_alias to reflect…

### LINKED ISSUE #653: lambda: `aws_lambda_alias` using the wrong Id format
### What happened?
When creating lambda aliases using provider-aws, the managed resource is created successfully, but then fails to maintain sync because the configured id function mapping doesn't match the id in tfstate.

### How can we reproduce it?
Create a lambda function and and alias to a published version. 
The alias will be provisioned in AWS, but the managed resource won't reconcile, as the external name will be updated incorrectly.
The provider is configured to use `function_name/alias` as the id, but tf uses the full ARN.

* Crossplane Version: 1.11.3-up.1
* Provider Version: v0.32.1
* Kubernetes Version: v1.25.3
* Kubernetes Distribution: kind

djeremiah:
… the true ID format

This PR changes the ExternalName configuration for aws_lambda_alias to match the id format expected by terraform. Specifically, the config was previously based on the import syntax for the resource, but TF uses the full ARN as the Id.

Fixes #653 

I have:

Local testing - provisioned a lambda function via S3 bucket and a function alias for a published version.

### DIFF
--- a/config/externalname.go
+++ b/config/externalname.go
@@ -965,8 +965,8 @@ var ExternalNameConfigs = map[string]config.ExternalName{
 
 	// lambda
 	//
-	// Lambda Function Aliases can be imported using the function_name/alias
-	"aws_lambda_alias": config.TemplatedStringAsIdentifier("name", "{{ .parameters.function_name }}/{{ .external_name }}"),
+	// Lambda Function Aliases are identified by their ARN, like arn:aws:lambda:eu-west-1:123456789012:function:lambda-function:alias
+	"aws_lambda_alias": config.TemplatedStringAsIdentifier("name", "arn:aws:lambda:{{ .setup.configuration.region }}:{{ .setup.client_metadata.account_id }}:function:{{ .parameters.function_name }}:{{ .external_name }}"),
 	// Code Signing Configs can be imported using their ARN that has a random
 	// substring in the end.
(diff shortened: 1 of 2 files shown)


jeanduplessis:
/test-examples="examples/lambda/alias.yaml"

jeanduplessis:
@djeremiah We have uptest failing on the current example in `examples/lambda/alias.yaml` – could you have a look to see if the example needs updating so we can ensure we keep e2e coverage for this resource.

djeremiah:
Do we generally want a single file to provide the e2e test? Or can we include other examples as dependencies? Testing the Alias will require publishing a lambda function. The plumbing for all of that appears to be fully included in examples/function.yaml, so we can either re-run the test including that example, or we can copy the needed items into alias.yaml.

jeanduplessis:
We tend to want to include all the necessary elements in the example file so that they can be run independently.

djeremiah:
/test-examples="examples/lambda/alias.yaml"

djeremiah:
/test-examples="examples/lambda/alias.yaml"

sergenyalcin (approved):
Thanks @djeremiah LGTM! I also tested and validated this PR in my local setup.

PR #680, opened 2023-04-26
Break provider-aws up by service

ulucinar:

Relevant proposal: https://github.com/crossplane/crossplane/pull/3939
Depends on: https://github.com/upbound/upjet/pull/194

This PR proposes the following changes according to the proposal https://github.com/crossplane/crossplane/pull/3939:
- Subpackages belonging to each API group is produced. An example is: provider-aws-kms.
- ProviderConfig, ProviderConfigUsage and StoreConfig are part of a config package named provider-aws-config.
- The monolith package (containing all the CRDs and associated controllers) is still produced.
- Each produced package except for the monolith package has the `pkg.crossplane.io/provider-family` label in its package metadata.
- Each service package except for the config package declares a dependency to the config package.

It also introduces the `SUBPACKAGES` make variable so that it's possible to build the individual subpackages. An example for building is:
```shell
make -j 3 SUBPACKAGES="monolith config ec2" build.all
```
One can also use the same variable to push those (already built) packages:
```shell
make -j 3 SUBPACKAGES="monolith config ec2" BRANCH_NAME=main publish
```

I have:

Successfully built and pushed the monolith, config and ec2 subpackages.

Also successfully provisioned and destroyed an `ec2.VPC` instance using the package `.../upbound/provider-aws-ec2:v0.34.0-rc.0.20.ga26202f1`.

### DIFF
--- a/package/crossplane-config.yaml
+++ b/package/crossplane-config.yaml
@@ -19,7 +19,3 @@ metadata:
       address. This is a subpackage for the {{ APIGROUP }} API group.
     friendly-name.meta.crossplane.io: Provider AWS ({{ APIGROUP }})
     auth.upbound.io/group: aws.upbound.io
-spec:
-  dependsOn:
-    - provider: {{ XPKG_REG_ORGS }}/provider-aws-config
-      version: ">={{ VERSION }}"
--- a/package/crossplane.yaml
+++ b/package/crossplane.yaml
@@ -22,4 +22,4 @@ metadata:
 spec:
   dependsOn:
     - provider: {{ XPKG_REG_ORGS }}/provider-aws-config
-      version: "*"
+      version: "{{ DEP_CONSTRAINT }}"
--- a/package/crossplane.yaml
+++ b/package/crossplane.yaml
@@ -22,4 +22,4 @@ metadata:
 spec:
   dependsOn:
     - provider: {{ XPKG_REG_ORGS }}/provider-aws-config
-      version: ">={{ VERSION }}"
+      version: "*"
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,8 @@ export TERRAFORM_DOCS_PATH ?= website/docs/r
 
 PLATFORMS ?= linux_amd64 linux_arm64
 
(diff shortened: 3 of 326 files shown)


package/crossplane-config.yaml
ulucinar, line 1: Do we know of a lightweight templating tool that we can quickly integrate into our build pipelines?

turkenh: We can use [gomplate](https://docs.gomplate.ca/). It is already available in the [build submodule](https://github.com/upbound/build/blob/292f958d2d97f26b450723998f82f7fc1767920c/makelib/local.mk#L82-L86) (we may want to move it out of local.mk though).

ulucinar:

package/crossplane-config.yaml
ulucinar: Thanks @turkenh. Let's proceed as separate manifests for now and once we have the installation of `gomplate` refactored into somewhere like the `k8s_tools.mk`, then we can utilize the tool for templating the `crossplane.yaml`.

package/crossplane.yaml
```diff
@@ -1,16 +1,25 @@
...
+      Available resources and their fields can be found in the [Upbound
+      Marketplace](https://marketplace.upbound.io/providers/upbound/provider-aws).
+      If you encounter an issue please reach out on support@upbound.io email
+      address. This is a subpackage for the {{ APIGROUP }} API group.
+    friendly-name.meta.crossplane.io: Provider AWS ({{ APIGROUP }})
     auth.upbound.io/group: aws.upbound.io
+spec:
+  dependsOn:
+    - provider: {{ XPKG_REG_ORGS }}/provider-aws-config
+      version: "*"
```
turkenh, line 25: Wouldn't this bring and rc version as a dependency to a released version? Which is why I suggested `>=v0.0.0` in the first place.

turkenh (approved):

ezgidemirel (approved):
I have updated the platform-ref-aws configuration with the sub-packages and successfully installed it. There are 160 CRDs installed to the cluster and I could successfully create the cluster claim.

PR #687, opened 2023-05-03
Use monolith for local run

turkenf:

While testing a new resource in provider-aws, after running `make run` in my local, the `ready` and `synced` status of the resource remained empty. I didn't encounter any error message in the description and when I used `monolith` package instead of `provider` in local run, the problem was solved.

Screenshot 2023-05-03 at 15 02 33

I have:


ulucinar:
Hi @turkenf,
Could you please fix the reported linter issue? It's been introduced in an earlier PR but because we had issues with the linter, we could not probably spot it.

ulucinar (approved):
Thanks @turkenf, lgtm.

PR #695, opened 2023-05-04
Add example to create cognito user pool with lambda triggers

### LINKED ISSUE #615: Cognito: `UserPool` CRD has incorrect restriction on lambdaConfig size
### What happened?

Cognito UserPool doesn't allow lambdaConfig to be created with more than one item, it errors out with 
```
Warning  CannotObserveExternalResource  3m31s (x4227 over 3d2h)  managed/cognitoidp.aws.upbound.io/v1beta1, kind=userpool  cannot run refresh: refresh failed: Too many lambda_config blocks: No more than 1 "lambda_config" blocks are allowed
```
Official Docs of Upbound say no such restriction [UpboundUserPoolDocs](https://marketplace.upbound.io/providers/upbound/provider-aws/v0.31.0/resources/cognitoidp.aws.upbound.io/UserPool/v1beta1)

Cognito UserPool has lambdaConfig which got translated from [Terraform definition](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/cognitoidp/user_pool.go#L231), and has an incorrect retriction of `MaxItems: 1`on the lambdaConfig items.

The official API documentation of AWS Cognito User Pool doesn't have such restriction and also has structure of Map.

User Pool Syntax: https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_CreateUserPool.html#API_CreateUserPool_RequestSyntax
Lambda Config Syntax: https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_LambdaConfigType.html

The Terraform provider somehow works around this to allow multiple lambda config elements as visible in their [example](https://github.com/hashicorp/terraform-provider-aws/blob/main/examples/cognito-user-pool/main.tf#L109)

### How can we reproduce it?

Create Cognito UserPool with more than one lambdaConfig Item 

* Universal Crossplane Version: v1.10.1
* Provider Version: v0.30.0

I also opened a bug card in [terraform-provider-aws](https://github.com/hashicorp/terraform-provider-aws/) [#30139](https://github.com/hashicorp/terraform-provider-aws/issues/30139)

thekaleidoscope:
* Add example to create cognito user pool with lambda triggers
* Add example to create image based lambda

This change will add an example showcasing how cognito user pool can setup lambda triggers, this is especially useful as the documentation requires lambdaConfig as array of key-value, but actually it expects an array with a single element with all the key-values in it.

Similarly, The Function (Lambda) doesn't have an example of using images to create lambda function, the change also includes an example of how it can be done as well.

Fixes #615

I have:

Similar example is used in our project, the example written was not directly tested as is.

Note(turkenf): Tested manually: https://github.com/upbound/provider-aws/pull/695#issuecomment-1584304396


turkenf:
I have successfully tested manually with the latest example manifest this resource:
```
NAME                                                 READY   SYNCED   EXTERNAL-NAME         AGE
userpool.cognitoidp.aws.upbound.io/lambda-example2   True    True     eu-west-1_JJgHQ4xGE   16m
```

PR is ready for review now @turkenh.

examples/cognitoidp/lambdatrigger.yaml
```diff
@@ -1,39 +0,0 @@
...
-    packageType: Image
-    image_uri: "123456789012.dkr.ecr.eu-west-2.amazonaws.com/example:0.1"
-    memorySize: 128
-    publish: true
-    region: eu-west-2
-    timeout: 5
-    roleSelector:
-      matchLabels:
-        testing.upbound.io/example-name: lambda-example
-
```
thekaleidoscope: an example of using container image in lamda also didn't exist, hence I thought this might be useful example too

examples/cognitoidp/userpool.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: cognitoidp.aws.upbound.io/v1beta1
+kind: UserPool
+metadata:
+  annotations:
+    meta.upbound.io/example-id: cognitoidp/v1beta1/userpool
```
turkenh, line 5: I belive we need manual intervention for this resource.

examples/cognitoidp/lambdatrigger.yaml
turkenf: Thank you for this PR @thekaleidoscope you are right, that would be useful. However, the `lambda.Function` resource is not relevant to this PR.  Also if we want to add an example of this we have to add it under the `lambda` group and the example name should contain the relevant resource, e.g. `function-with-container-image.yaml`.

Feel free to open a new PR adding an example for `lambda.Function` with container image.

examples/cognitoidp/userpool.yaml
turkenf: I have added manual intervention.

turkenh (approved):

PR #699, opened 2023-05-14
Call `up xpkg batch` to batch-process the smaller provider packages in the build pipelines

ulucinar:

Relevant PR: https://github.com/upbound/provider-gcp/pull/299

This PR invokes the `up xpkg batch` command to streamline the processing of the smaller provider packages. The `xpkg build` implementation dumps the built packages onto the file system for `xpkg push` to load them back from the filesystem and push them to a package registry. The `xpkg batch` command does not need to store the built provider packages in the filesystem. The packages are instead directly pushed to the configured registry. 

All smaller packages built by `xpkg batch` share a common base image and the `img.build` make target is modified so that it now builds only this common base image instead of the individual smaller provider images. `xpkg batch` adds the `base` and the `upbound` layers as before, as well as a layer containing the provider controller binary. The `base` and `upbound` labeled layers are shared between all the architectures of a provider's package.

`xpkg batch` can also build & push the provider packages in parallel with the configured (via the `--concurrency` option) maximum concurrency. 

`xpkg batch` also has a templating engine so that the family of smaller providers being built can share a template provider package metadata. The templating engine substitutes the `{{ .Service }}` and `{{ .Name }}` template variables with the service and package repo names, respectively, and further template variables can also be supplied via multiple `--template-var` command-line options. This template metadata resides at `package/crossplane.yaml.tmpl` and is introduced with this PR. 

This PR also bumps `up` to `v0.17.0`.

I have:

Uptest run: https://github.com/upbound/provider-aws/actions/runs/4974824490
Pipeline run with a subset of smaller providers (config, ec2) and the monolith: https://github.com/upbound/provider-aws/actions/runs/4979367740


ulucinar:

sergenyalcin (approved):
This is already reviewed in https://github.com/upbound/provider-gcp/pull/299. Thanks @ulucinar LGTM!

PR #703, opened 2023-05-17
Fix for the issue 574: rds: DBCluster writeConnectionSecret missing fields

svscheg:
Add next fields: write the username and port field to the connection secret
image

Fixes https://github.com/upbound/provider-aws/issues/574

I have:

Uptest: https://github.com/upbound/provider-aws/actions/runs/5212568558
Manually
image

### DIFF
--- a/config/rds/config.go
+++ b/config/rds/config.go
@@ -52,6 +52,12 @@ func Configure(p *config.Provider) {
 			if a, ok := attr["reader_endpoint"].(string); ok {
 				conn["reader_endpoint"] = []byte(a)
 			}
+			if a, ok := attr["master_username"].(string); ok {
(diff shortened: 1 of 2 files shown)


turkenh (approved):
Thank you @svscheg!

Noticed that we didn't trigger uptest here and the resource has a manual intervention which I believe we could avoid by adding a secret. 

It would be nice if we could use this as a chance to automate and remove that manual intervention and so that we can test this resource with uptest.

svscheg:
/test-examples="examples/rds/cluster.yaml"

svscheg:
Thank you, @turkenh !
Deleted manual-intervention and added a secret, now we can run it using uptest.

PR #707, opened 2023-05-19
feat(firehose): add hec_token as sensitive

haarchri:
set hec_token as sensitive

Fixes #

I have:

create, update, delete


ytsarev (commented):
looks good! Can we cover it with uptest example?

turkenf (approved):
> looks good! Can we cover it with uptest example?

@ytsarev, I believe that this is not a generic enough scenario to automate. Also, here we will need to create [splunk configuration](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/kinesis_firehose_delivery_stream#splunk-destination:~:text=%3D%20%22splunk%22-,splunk_configuration,-%7B%0A%20%20%20%20hec_endpoint%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%3D) block to create an example with `hec_token`, which would most probably require manual intervention.

So I think we can rely on @haarchri's manual testing and merge this PR.

PR #747, opened 2023-06-21
fix(iam): Policy ID should contain path and the external-name derived from the ID should be the name part only

### LINKED ISSUE #742: Incorrect ID for policies with paths
### What happened?
After restarting the AWS provider, `policy` objects with a path become not `Ready`. I believe it is because the initial tfstate id and the composed tfstate id are different. The composed state after a restart does not contain the path in the ARN used in the id field. This results in an error:

```
apply failed: creating IAM Policy (crossplane-app-policy-pg-license-only-test):
      EntityAlreadyExists: A policy called crossplane-app-policy-pg-license-only-test
      already exists. Duplicate names are not allowed.\n\tstatus code: 409, request
      id: uuid-of-error-event
```

### How can we reproduce it?
Create an IAM policy object with a path and restart the IAM provider. 

* Crossplane Version: v1.12.2
* Provider Version: v0.36.0
* Kubernetes Version: 1.26

portswigger-tim:

Terraform uses the policy path in the `tfstate` id, and name (derived from the ID) must match [\w+=,.@-]. Both are bugs in v0.36.0, both are fixed in this PR.

Fixes #742 

I have:

Tested by running locally against existing non-ready resources and by creating new resources with and without paths.

### DIFF
--- a/config/externalname.go
+++ b/config/externalname.go
@@ -486,7 +486,7 @@ var ExternalNameConfigs = map[string]config.ExternalName{
 	"aws_iam_access_key":       config.IdentifierFromProvider,
 	"aws_iam_instance_profile": config.NameAsIdentifier,
 	// arn:aws:iam::123456789012:policy/UsersManageOwnCredentials
-	"aws_iam_policy": config.TemplatedStringAsIdentifier("name", "arn:aws:iam::{{ .setup.client_metadata.account_id }}:policy/{{ .external_name }}"),
+	"aws_iam_policy": iamPolicy(),
 	"aws_iam_user":   config.NameAsIdentifier,
 	"aws_iam_group":  config.NameAsIdentifier,
 	"aws_iam_role":   config.NameAsIdentifier,
@@ -2636,6 +2636,26 @@ func iamUserGroupMembership() config.ExternalName {
 		}
 		return strings.Join(append([]string{u.(string)}, groups...), "/"), nil
 	}
+	return e
+}
+
+func iamPolicy() config.ExternalName {
+	e := config.TemplatedStringAsIdentifier("name", "arn:aws:iam::{{ .setup.client_metadata.account_id }}:policy{{ .parameters.path }}{{ .external_name }}")
+	e.GetExternalNameFn = func(tfstate map[string]any) (string, error) {
+		id, ok := tfstate["id"]
+		if !ok {
+			return "", errors.New("id attribute missing from state file")
+		}
+
+		idStr, ok := id.(string)
+		if !ok {
+			return "", errors.New("value of id needs to be string")
+		}
+
+		arnSlice := strings.Split(idStr, "/")
+		return arnSlice[len(arnSlice)-1], nil
+	}
+
 	return e
 }
 
--- a/config/externalname.go
+++ b/config/externalname.go
@@ -2640,7 +2640,17 @@ func iamUserGroupMembership() config.ExternalName {
 }
 
 func iamPolicy() config.ExternalName {
-	e := config.TemplatedStringAsIdentifier("name", "arn:aws:iam::{{ .setup.client_metadata.account_id }}:policy{{ .parameters.path }}{{ .external_name }}")
+	e := config.NameAsIdentifier
+
+	e.GetIDFn = func(_ context.Context, externalName string, parameters map[string]any, setup map[string]any) (string, error) {
+		path, ok := parameters["path"]
+		if !ok {
+			path = "/"
+		}
+		accountID := setup["client_metadata"].(map[string]string)["account_id"]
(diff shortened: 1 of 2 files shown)


julienduchesne:
We have integrated this fix in our clusters and it works great! We have ~20 policies with paths and they have all fully resolved

portswigger-tim:
Thanks for trying it out @julienduchesne - Hopefully this PR makes its way into the next release 😄

jeanduplessis:
/test-examples="examples/iam/policy.yaml"

jeanduplessis:
@portswigger-tim It seems that `examples/iam/policy.yaml` is failing with this update when running the end to end test with Uptest.. 
Could you have a look at the example file to see if anything needs changing or you could have a look at the data from the test run to see why it might be failing: https://github.com/upbound/provider-aws/suites/13796558139/artifacts/765074641

sergenyalcin:
In the test logs there is the following error:

```
events	cannot run refresh: refresh failed: reading IAM Policy (arn:aws:iam::153891904029:policy<no value>sample-user-policy): InvalidInput: ARN arn:aws:iam::153891904029:policy&lt;no value&gt;sample-user-policy is not valid.
	status code: 400, request id: 70bb2c3e-ee97-4268-b3f2-51da4a09cd80: 	{"type": "Warning", "object": {"kind":"Policy","name":"sample-user-policy","uid":"7c6d5a44-3982-402c-b07a-2a6d2a878208","apiVersion":"iam.aws.upbound.io/v1beta1","resourceVersion":"7553"}, "reason": "CannotObserveExternalResource"}
```

For this case, the constructed id is:

`arn:aws:iam::153891904029:policy<no value>sample-user-policy`

As I see, the template of ID of this change is:

`config.TemplatedStringAsIdentifier("name", "arn:aws:iam::{{ .setup.client_metadata.account_id }}:policy{{ .parameters.path }}{{ .external_name }}")`

It seems that, the `.parameters.path` is empty in the used example. So while filling this field in the example file we can fix this error.

However, I think there is an important problem here. The [path field](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_policy#path) has a default but also this is an `optional` one and with the new configuration we are relying to an `Optional` field. In this case, crossplane has not any defaulting ability. So, while constructing ID, if the `path` field is empty, provider cannot construct the ID and we cannot create resource.

At this point I can have a suggestion for fixing the issue by a tricky defaulting in configuration path:

```go
func iamPolicy() config.ExternalName {
	e := config.IdentifierFromProvider
	e.GetIDFn = func(_ context.Context, externalName string, parameters map[string]any, setup map[string]any) (string, error) {
		path, ok := parameters["path"]
		if !ok {
			path = "/"
		} else {
			path = fmt.Sprintf("%s/", path)
		}
		accountID := setup["client_metadata"].(map[string]string)["account_id"]
		return fmt.Sprintf("arn:aws:iam::%s:policy%s%s", accountID, path, externalName), nil
	}
	e.GetExternalNameFn = func(tfstate map[string]any) (string, error) {
		id, ok := tfstate["id"]
		if !ok {
			return "", errors.New("id attribute missing from state file")
		}

		idStr, ok := id.(string)
		if !ok {
			return "", errors.New("value of id needs to be string")
		}

		arnSlice := strings.Split(idStr, "/")
		return arnSlice[len(arnSlice)-1], nil
	}

	return e
}
```

I did not test it but the main goal is here, checking the value of `path` field and according to the value, constructing the ID. As I mentioned, this suggestion should be tested in your cases and also in Uptest side.

portswigger-tim:
/test-examples="examples/iam/policy.yaml"

portswigger-tim:
Thanks for the feedback and hints @sergenyalcin - I've amended `iamPolicy()`, I've also created a file: `"examples/iam/policy-with-path.yaml"` for testing.

I can confirm that running locally works against pre-existing and new policy resources with and without the path parameter set.

I have chosen not to fmt the path with a slash in it as AWS documents that: 
> This parameter allows (through its [regex pattern](http://wikipedia.org/wiki/regex)) a string of characters consisting of either a forward slash (/) by itself or a string that must begin and end with forward slashes. In addition, it can contain any ASCII character from the ! (\u0021) through the DEL character (\u007F), including most punctuation characters, digits, and upper and lowercased letters.

This is generally how you specify paths for roles, policies etc in all AWS IaC tools like Terraform and CloudFormation. Allowing users to not include the leading and trailing slash might cause confusion.

jeanduplessis:
/test-examples="examples/iam/policy.yaml"

jeanduplessis:
/test-examples="examples/iam/policy-with-path.yaml"

sergenyalcin (approved):
Thanks @portswigger-tim LGTM!

PR #789, opened 2023-07-14
Add an event filter with the resource.DesiredStateChanged predicate to filter status updates out

ulucinar:

This PR adds an event filter using the [resource.DesiredStateChanged](https://github.com/crossplane/crossplane-runtime/blob/27a547d7e4fc71f4ecf8cbe155a820a283ae6c0e/pkg/resource/predicates.go#L170) to the generated resource controllers so that updates to only labels, annotations (Crossplane's create pending/failed annotations will be ignored) and to the `spec` will result in reconcile requests.

I have:

Tested with https://github.com/upbound/upjet/pull/231.


config/kendra/config.go
```diff
@@ -0,0 +1,26 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kendra
+
+import (
+	"github.com/upbound/upjet/pkg/config"
+)
+
+// Configure adds configurations for kafka group.
```
sergenyalcin, line 21:
```suggestion
// Configure adds configurations for kendra group.
```

config/medialive/config.go
```diff
@@ -0,0 +1,26 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package medialive
+
+import (
+	"github.com/upbound/upjet/pkg/config"
+)
+
+// Configure adds configurations for kafka group.
```
sergenyalcin, line 21:
```suggestion
// Configure adds configurations for medialive group.
```

sergenyalcin (approved):
Thanks @ulucinar I left two comments that are related to copy-paste.

config/kendra/config.go
ulucinar: Thanks @sergenyalcin, done.

config/medialive/config.go
ulucinar: Done.

PR #815, opened 2023-08-04
Update KnownReferencers to apply to non-top-level schema elements

### LINKED ISSUE #803: kmsKeyArnSelector

The KMS key for `cluster.eks.aws.upbound.io` `spec.forProvider.encryptionConfig` requires the KMS key ARN.

https://marketplace.upbound.io/providers/upbound/provider-aws-eks/v0.37.0/resources/eks.aws.upbound.io/Cluster/v1beta1#doc:spec-forProvider-encryptionConfig-provider-keyArn

I want to create a `key.kms.aws.upbound.io` and `cluster.eks.aws.upbound.io` within the same Composition and use the KMS key for the cluster encryption config.

In order to facilitate creating a KMS key and using it for the Cluster config at the same time, it would be useful to implement a `kmsKeyArnSelector`.

This would be similar to the `kmsKeyArnRef` implemented for `seccurityconfiguration.glue.aws.upbound.io` (https://marketplace.upbound.io/providers/upbound/provider-aws/v0.37.0/resources/glue.aws.upbound.io/SecurityConfiguration/v1beta1)

### Workaround:

My current workaround is to store the KMS key ARN as a field on the XR status populated by a patch:
```
        - type: ToCompositeFieldPath
          fromFieldPath: status.atProvider.arn
          toFieldPath: status.clusterKmsKeyArn
          policy:
            fromFieldPath: Required
```

And then patch that into the Cluster config:
```
       - type: FromCompositeFieldPath
          fromFieldPath: status.clusterKmsKeyArn
          toFieldPath: spec.forProvider.encryptionConfig[0].provider[0].keyArn
          policy:
            fromFieldPath: Required
```

mbbush:

The `KnownReferencers` function uses a heuristic to find fields that look like they probably reference other common resources, saving the effort of manually specifying every one of them. However, we still frequently see requests like "the such-and-such kmsKeyId field is missing a selector", because `KnownReferencers` only looks at the top level fields in the schema. 

I've updated it to recurse down the schema tree and also inspect child fields. I also added a few more commonly referenced fields, and made the kms ones broader (suffix rather than exact match)

This is my second time writing go (first was https://github.com/crossplane/crossplane/pull/4371). I'm sure there's room for improvement here, and I would appreciate suggestions for improvements.

Fixes #803

I have:

I was unable to run the linter because of the known memory issues. It OOMs after consuming over 30GB of ram.

One of the main reasons I'm opening this as a draft is that I'm hoping for some guidance on how best to test something like this. I'm a little fuzzy on how the testing using the `examples` and `generated-examples` works, and would appreciate either some explanation or a link to some documentation. 

I've prepared lists of resources with inferred references added, (from the `Println` statements which I will either remove or comment out before merging) from each of the various commits, including my early refactor commits which I confirmed do not change any files when I run `make generate`. I'll post them in a PR comment.

I did not commit the results of `make generate` until the final commit in the PR.


mbbush:
Before this change, `KnownReferencers` generated [227 references](https://github.com/upbound/provider-aws/files/12262987/before-my-change.7db8f3e.txt).

When I updated the `kms` fields to be a suffix rather than a full match, there were [9 more references for a total of 236](https://github.com/upbound/provider-aws/files/12262988/kms-as-a-suffix.77280b6.txt) added

Adding recursion made a big difference. Now there are a total of [428 references of various types](https://github.com/upbound/provider-aws/files/12262985/add-recursion.77c03b8.txt)

Adding Function.lambda, Queue.sqs, and Group.cloudwatchlogs brought the [total to 471](https://github.com/upbound/provider-aws/files/12262984/add-lambda-sqs-log-group.c63b690.txt)

Adding Bucket.s3 brought the [total to 507](https://github.com/upbound/provider-aws/files/12262986/add-s3-bucket.fab43a4.txt)

mbbush:
I think it's probably better to just have more of these references specified explicitly, rather than inferring a bunch, so I'm going to close this.

PR #896, opened 2023-09-27
Configure a default poll jitter for the controllers

ulucinar:

This PR proposes configuring a default poll jitter of 5% of the poll interval for up-to-date resources so that if the reconciliations of the resources are synchronized for some reason (e.g., pod restart), they will not be queued for reconciliation at the same moment. Also as the provider keeps reconciling the resources, the reconciliations will be randomly distributed. We initially start with a poll jitter of 5% of the poll interval and currently this interval is not configurable via the command-line options not to increase the configuration complexity. If there's need to configure the jitter for different environments, then we can add this option in a future PR. 

The configured jitter is also dumped in the debug logs when the provider is starting:
```
2023-09-27T22:10:33+03:00       DEBUG   provider-aws    Starting        {"sync-interval": "1h0m0s", "poll-interval": "10m0s", "poll-jitter": "30s", "max-reconcile-rate": 10}
```

I have:

Tested manually with a `Role.iam` resource.


ulucinar:

turkenf (approved):
Thank you @ulucinar, LGTM.

PR #913, opened 2023-10-09
Bump uptest to v0.6.0

ulucinar:

- This activates the resource import and resource update tests.
- For now, we don't specify a default update parameter at the provider level, and resource update tests will currently need to be enabled with the "uptest.upbound.io/update-parameter" annotation in the example manifest on a per-resource basis.
- `uptest` now also has more verbose logging to make troubleshooting uptest runs a little bit easier :)

Resource import tests make sure that the external resource state can be re-imported in case the local state is lost, for instance, due to a pod restart and the managed resource stays in `Synced == True, Ready == True` state. The optional resource update tests update a designated `spec.forProvider` field (the desired state) and assert that the updated value becomes available at `status.atProvider` (the actual state) and the managed resource stays at the `Synced == True, Ready == True` state. The `spec.forProvider` field to be used in the update tests for a specific resource can be specified with the `uptest.upbound.io/update-parameter` in the resource's example manifest to be tested with `uptest`:

```yaml
...
metadata:
  annotations:
    uptest.upbound.io/update-parameter: '{"tags":{"update-test-tag":"val"}}'
```

I have:

Manually tested with the `examples/ec2/vpc.yaml` manifest.


turkenf (approved):
Thank you @ulucinar, LGTM.

PR #920, opened 2023-10-11
Add reference to permission set to ssoadmin AccountAssignment

mbbush:
### Description of your changes

Add a reference to the ssoadmin PermissionSet to ssoadmin AccountAssignment 

Fixes #888 

I have:

- [ ] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

I'm planning a larger PR to make working with SSO resources nicer (adding some missing resources and references to them). 
Because these examples require manual intervention, I'll run the tests locally against my aws account which has a real SSO instance and post the results.

PR #928, opened 2023-10-20
Add full support for SSO resources

### LINKED ISSUE #536: Moving ssoadmin(3), identitystore(3), inspector2(2) resources to v1beta1 version
* `ssoadmin`
  * [x] [aws_ssoadmin_customer_managed_policy_attachment](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ssoadmin_customer_managed_policy_attachment)
  * [x] [aws_ssoadmin_instance_access_control_attributes](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ssoadmin_instance_access_control_attributes)
  * [x] [aws_ssoadmin_permissions_boundary_attachment](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ssoadmin_permissions_boundary_attachment)

* `identitystore`
  * [x] [aws_identitystore_group](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/identitystore_group)
  * [x] [aws_identitystore_group_membership](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/identitystore_group_membership)
  * [x] [aws_identitystore_user](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/identitystore_user)

* `inspector2`
  * [ ] [aws_inspector2_delegated_admin_account](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/inspector2_delegated_admin_account)
  * [ ] [aws_inspector2_organization_configuration](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/inspector2_organization_configuration)

### LINKED ISSUE #888: AccountAssignment.ssoadmin.aws.upbound.io/v1beta1 is missing permissionSetArnRef.

We're using `ssoadmin.aws.upbound.io/v1beta1` for our AWS multi-account IAM set up and while most of the CRDs support `permissionSetArnRef`, the `AccountAssignment` spec is missing it, which requires either hardcoding the ARN in or a patch.  It'd be great if we just had parity across all those CRDs so it could be fully dynamic and reusable.

https://github.com/upbound/provider-aws/blob/main/package/crds/ssoadmin.aws.upbound.io_permissionsetinlinepolicies.yaml#L83 is an example of the field I'm referring to.

https://github.com/upbound/provider-aws/blob/main/package/crds/ssoadmin.aws.upbound.io_accountassignments.yaml#L328 is where it's omitted.

mbbush:
* Added 3 resources in the identitystore api group, which had been completely missing (these are the users and groups used for single sign on)
* Added 3 missing resources to the ssoadmin api group, for full support, with external names as normalized as possible
* Normalized the external names of two existing ssoadmin resources (PermissionSet and PermissionSetInlinePolicy)
* Added cross references
* Tested everything and verified it works.
* Rebased on the latest `main`

This incorporates the changes in #920 and has been tested, so I'll close #920.

Fixes #536 (mostly)
Fixes #888
Closes #920 

I have:

- [x] Run `make check-diff test` to validate the unit tests and generated code. Running the linter causes out-of-memory errors (it was using more than 50GB).

* At commit aab2f0d547f06f57636736852095a886acbc91f9, using uptest locally:
I removed the manual-intervention annotations, substituted in my account-specific values, and ran  `make uptest` with all the examples in this PR. All the tests passed (create, update, import, delete). 

I did discover one breaking change from the switch to the no-fork provider. Details at #1013 

Here's a snapshot of `while true; do date; kubectl get managed; done` from the time the first uptest step completed, showing everything synced and ready.
```
Wed Dec  6 12:24:33 PM PST 2023
NAME                                                                    READY   SYNCED   EXTERNAL-NAME                                 AGE
policy.iam.aws.upbound.io/ssoadmin-customer-managed-policy-attachment   True    True     ssoadmin-customer-managed-policy-attachment   29s
policy.iam.aws.upbound.io/ssoadmin-permissions-boundary-attachment      True    True     ssoadmin-permissions-boundary-attachment      29s

NAME                                                              READY   SYNCED   EXTERNAL-NAME                          AGE
groupmembership.identitystore.aws.upbound.io/example-membership   True    True     a4e884d8-c001-700a-f74a-a59b8050ed2e   29s

NAME                                                            READY   SYNCED   EXTERNAL-NAME                          AGE
group.identitystore.aws.upbound.io/example-group                True    True     74b8c4f8-60f1-7054-fe80-959d4c8a48c8   29s
group.identitystore.aws.upbound.io/example-membership           True    True     0478f4f8-3061-70cc-1fba-d9bc9bdebfe8   29s
group.identitystore.aws.upbound.io/ssoadmin-accountassignment   True    True     44d8e408-10a1-7076-de4e-eb531a66e562   29s

NAME                                                   READY   SYNCED   EXTERNAL-NAME                          AGE
user.identitystore.aws.upbound.io/example-membership   True    True     347884b8-3071-70ae-cf6d-236109cfcc9a   29s
user.identitystore.aws.upbound.io/example-user         True    True     74a85488-a021-70f6-cb73-f25adb5bb918   29s

NAME                                                                                 READY   SYNCED   EXTERNAL-NAME                                                                                                                                                                AGE
managedpolicyattachment.ssoadmin.aws.upbound.io/ssoadmin-managed-policy-attachment   True    True     arn:aws:iam::aws:policy/AlexaForBusinessDeviceSetup,arn:aws:sso:::permissionSet/ssoins-7223ae4e1f24fd4c/ps-cbcf5801ea90aec7,arn:aws:sso:::instance/ssoins-7223ae4e1f24fd4c   29s

NAME                                                                                                  READY   SYNCED   EXTERNAL-NAME                                    AGE
instanceaccesscontrolattributes.ssoadmin.aws.upbound.io/ssoadmin-instance-access-control-attributes   True    True     arn:aws:sso:::instance/ssoins-7223ae4e1f24fd4c   29s

NAME                                                                                                  READY   SYNCED   EXTERNAL-NAME                                                                                                                                                          AGE
customermanagedpolicyattachment.ssoadmin.aws.upbound.io/ssoadmin-customer-managed-policy-attachment   True    True     ssoadmin-customer-managed-policy-attachment,/,arn:aws:sso:::permissionSet/ssoins-7223ae4e1f24fd4c/ps-4a48766aef623884,arn:aws:sso:::instance/ssoins-7223ae4e1f24fd4c   29s

NAME                                                                                             READY   SYNCED   EXTERNAL-NAME         AGE
permissionsboundaryattachment.ssoadmin.aws.upbound.io/ssoadmin-permissions-boundary-attachment   True    True     ps-3d4a4ee185ba4972   29s

NAME                                                                                READY   SYNCED   EXTERNAL-NAME                                                                                                            AGE
permissionset.ssoadmin.aws.upbound.io/ssoadmin-accountassignment                    True    True     arn:aws:sso:::permissionSet/ssoins-7223ae4e1f24fd4c/ps-a191ea70e4f281d0,arn:aws:sso:::instance/ssoins-7223ae4e1f24fd4c   29s
permissionset.ssoadmin.aws.upbound.io/ssoadmin-customer-managed-policy-attachment   True    True     arn:aws:sso:::permissionSet/ssoins-7223ae4e1f24fd4c/ps-4a48766aef623884,arn:aws:sso:::instance/ssoins-7223ae4e1f24fd4c   29s
permissionset.ssoadmin.aws.upbound.io/ssoadmin-managed-policy-attachment            True    True     arn:aws:sso:::permissionSet/ssoins-7223ae4e1f24fd4c/ps-cbcf5801ea90aec7,arn:aws:sso:::instance/ssoins-7223ae4e1f24fd4c   29s
permissionset.ssoadmin.aws.upbound.io/ssoadmin-permission-set                       True    True     arn:aws:sso:::permissionSet/ssoins-7223ae4e1f24fd4c/ps-0dd46224d50dba40,arn:aws:sso:::instance/ssoins-7223ae4e1f24fd4c   29s
permissionset.ssoadmin.aws.upbound.io/ssoadmin-permission-set-inline-policy         True    True     arn:aws:sso:::permissionSet/ssoins-7223ae4e1f24fd4c/ps-c85d97c97779dcc4,arn:aws:sso:::instance/ssoins-7223ae4e1f24fd4c   29s
permissionset.ssoadmin.aws.upbound.io/ssoadmin-permissions-boundary-attachment      True    True     arn:aws:sso:::permissionSet/ssoins-7223ae4e1f24fd4c/ps-3d4a4ee185ba4972,arn:aws:sso:::instance/ssoins-7223ae4e1f24fd4c   29s

NAME                                                                   READY   SYNCED   EXTERNAL-NAME                                                                                                                                                                                AGE
accountassignment.ssoadmin.aws.upbound.io/ssoadmin-accountassignment   True    True     44d8e408-10a1-7076-de4e-eb531a66e562,GROUP,374300188748,AWS_ACCOUNT,arn:aws:sso:::permissionSet/ssoins-7223ae4e1f24fd4c/ps-a191ea70e4f281d0,arn:aws:sso:::instance/ssoins-7223ae4e1f24fd4c   29s

NAME                                                                                      READY   SYNCED   EXTERNAL-NAME                                                                                                            AGE
permissionsetinlinepolicy.ssoadmin.aws.upbound.io/ssoadmin-permission-set-inline-policy   True    True     arn:aws:sso:::permissionSet/ssoins-7223ae4e1f24fd4c/ps-c85d97c97779dcc4,arn:aws:sso:::instance/ssoins-7223ae4e1f24fd4c   29s

```

### How can someone else test this code?
AWS SSO can only be enabled in the "management account" of an organization. The easiest way to do this is in a standalone personal account. You can go to SSO identity center in the aws console to enable this, which will allow you to make your account into an organization. This will allow you to generate both an "identity store id" and an "sso instance arn" which are the two parameters you need to manually supply to the identitystore and ssoadmin example resources, respectively. 

The example resources must be created in the same region as your SSO instance. I made my SSO instance in us-east-1, so that's the region I committed to all the example manifests. 

AWS does not charge for any of these SSO resources, so there were no financial considerations to testing with a personal account.


examples/ssoadmin/customermanagedpolicyattachment.yaml
```diff
@@ -0,0 +1,66 @@
+apiVersion: ssoadmin.aws.upbound.io/v1beta1
+kind: CustomerManagedPolicyAttachment
+metadata:
+  annotations:
+    meta.upbound.io/example-id: ssoadmin/v1beta1/customermanagedpolicyattachment
```
turkenf, line 5: Could you please add a manual intervention like below?

```suggestion
    meta.upbound.io/example-id: ssoadmin/v1beta1/customermanagedpolicyattachment
    upjet.upbound.io/manual-intervention: "This resource depends on PermissionSet. This resource is skipping because the PermissionSet resource was skipped."
```

examples/ssoadmin/customermanagedpolicyattachment.yaml
```diff
@@ -0,0 +1,66 @@
...
+        testing.upbound.io/example-name: ssoadmin-customer-managed-policy-attachment
+    region: us-east-1
+
+---
+
+apiVersion: iam.aws.upbound.io/v1beta1
+kind: Policy
+metadata:
+  annotations:
+    meta.upbound.io/example-id: ssoadmin/v1beta1/customermanagedpolicyattachment
```
turkenf, line 31:
```suggestion
    meta.upbound.io/example-id: ssoadmin/v1beta1/customermanagedpolicyattachment
    upjet.upbound.io/manual-intervention: "This resource is skipping because the parent resource was skipped."
```

examples/ssoadmin/managedpolicyattachment.yaml
```diff
@@ -3,18 +3,18 @@ kind: ManagedPolicyAttachment
...
+    upjet.upbound.io/manual-intervention: "This resource requires a valid instanceArn(The Amazon Resource Name (ARN) of the SSO Instance under which
     the operation will be executed)."
   labels:
-    testing.upbound.io/example-name: example
-  name: example
+    testing.upbound.io/example-name: ssoadmin-managed-policy-attachment
+  name: ssoadmin-managed-policy-attachment
 spec:
   forProvider:
     instanceArn: ${tolist(data.aws_ssoadmin_instances.example.arns)[0]}
```
turkenf, line 13: nit: for consistency, it may be better to put the same value on all examples, either we should adopt other examples to this, or we should adapt these others.

```suggestion
    instanceArn: ARN_OF_MANUALLY_CREATED_SSO_INSTANCE
```

examples/ssoadmin/managedpolicyattachment.yaml
```diff
@@ -27,10 +27,10 @@ metadata:
     upjet.upbound.io/manual-intervention: "This resource requires a valid instanceArn(The Amazon Resource Name (ARN) of the SSO Instance under which 
     the operation will be executed)."
   labels:
-    testing.upbound.io/example-name: example
-  name: example
+    testing.upbound.io/example-name: ssoadmin-managed-policy-attachment
+  name: ssoadmin-managed-policy-attachment
 spec:
   forProvider:
     instanceArn: ${tolist(data.aws_ssoadmin_instances.example.arns)[0]}
```
turkenf, line 34:
```suggestion
    instanceArn: ARN_OF_MANUALLY_CREATED_SSO_INSTANCE
```

examples/ssoadmin/permissionsboundaryattachment.yaml
```diff
@@ -0,0 +1,66 @@
+apiVersion: ssoadmin.aws.upbound.io/v1beta1
+kind: PermissionsBoundaryAttachment
+metadata:
+  annotations:
+    meta.upbound.io/example-id: ssoadmin/v1beta1/permissionsboundaryattachment
```
turkenf, line 5: Could you please add a manual intervention?

examples/ssoadmin/permissionsboundaryattachment.yaml
```diff
@@ -0,0 +1,66 @@
...
+        path: /
+    region: us-east-1
+
+---
+
+apiVersion: iam.aws.upbound.io/v1beta1
+kind: Policy
+metadata:
+  annotations:
+    meta.upbound.io/example-id: ssoadmin/v1beta1/permissionsboundaryattachment
```
turkenf, line 31: Could you please add a manual intervention?

examples/ssoadmin/permissionset.yaml
```diff
@@ -6,13 +6,13 @@ metadata:
...
     the operation will be executed)."
   labels:
-    testing.upbound.io/example-name: example
-  name: example
+    testing.upbound.io/example-name: ssoadmin-permission-set
+  name: ssoadmin-permission-set
 spec:
   forProvider:
     description: An example
     instanceArn: ${tolist(data.aws_ssoadmin_instances.example.arns)[0]}
```
turkenf, line 14:
```suggestion
    instanceArn: ARN_OF_MANUALLY_CREATED_SSO_INSTANCE
```

config/ssoadmin/config.go
```diff
@@ -0,0 +1,36 @@
...
+	})
+	p.AddResourceConfigurator("aws_ssoadmin_customer_managed_policy_attachment", func(r *config.Resource) {
+		r.References["customer_managed_policy_reference.name"] = config.Reference{
+			TerraformName:     "aws_iam_policy",
+			RefFieldName:      "PolicyNameRef",
+			SelectorFieldName: "PolicyNameSelector",
+		}
+	})
+	p.AddResourceConfigurator("aws_ssoadmin_permission_set_inline_policy", func(r *config.Resource) {
+		r.References["instance_arn"] = config.Reference{
```
turkenf, line 31: I guess this reference should be defined for `permission_set_arn`

```suggestion
		r.References["permission_set_arn"] = config.Reference{
```

examples/ssoadmin/permissionsetinlinepolicy.yaml
```diff
@@ -6,21 +6,22 @@ metadata:
...
-    testing.upbound.io/example-name: example
-  name: example
+    testing.upbound.io/example-name: ssoadmin-permission-set-inline-policy
+  name: ssoadmin-permission-set-inline-policy
 spec:
   forProvider:
-    instanceArn: ${tolist(data.aws_ssoadmin_instances.example.arns)[0]}
+    instanceArnSelector:
+      matchLabels:
+        testing.upbound.io/example-name: ssoadmin-permission-set-inline-policy
```
turkenf, line 15: We may need to remove these lines after reference verification?

turkenf (commented):
Thank you for your effort @mbbush, I left a few comments. Could you please add screenshots/logs of all resources to the section on how it was tested after making the changes, as [here](https://github.com/upbound/provider-aws/pull/514#issue-1582964680)?

examples/ssoadmin/customermanagedpolicyattachment.yaml
mbbush: I think there's a missing piece in my understanding of what these annotations are used for.

In my mental model, the goal of this annotation is to avoid creating managed resources that are in a state which will never become ready, because there is some additional information needed to create them. The user running the test needs to manually create them after the test has started, and then they will become ready, and the test will finish and pass. In that model, it seems like it would be better to _not_ have a manual intervention annotation on this resource, because if it's applied as is, it will simply wait for the permissionset to be (manually) created and ready, thanks to the selector, and then it will become ready without manual intervention needed. 

If we put the manual intervention annotation on, it just makes more manual work for the user running the test.

What am I overlooking/misunderstanding?

turkenf: You are correct when we consider it from your mental model, but the simple purpose of placing manual intervention is skipping the automatic test. The aim is to provide a brief explanation for the reason behind the skip, ensuring that the person triggering the test is informed.

In this case, a person who triggers an automatic test(uptest) on the `CustomerManagedPolicyAttachment` resource can see why it was skipped with this annotations `upjet.upbound.io/manual-intervention: "This resource depends on PermissionSet. This resource is skipping because the PermissionSet resource was skipped."`

config/ssoadmin/config.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+// Configure adds configurations for the ssoadmin group.
+func Configure(p *config.Provider) {
+	p.AddResourceConfigurator("aws_ssoadmin_account_assignment", func(r *config.Resource) {
+		r.References["principal_id"] = config.Reference{
+			TerraformName:     "aws_identitystore_group",
+			Extractor:         `github.com/crossplane/upjet/pkg/resource.ExtractParamPath("group_id",true)`,
+			RefFieldName:      "PrincipalGroupRef",
+			SelectorFieldName: "PrincipalGroupSelector",
```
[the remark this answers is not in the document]
mbbush: The reason I put the word group in the field name is to indicate that the selector (unlike the field) will only work if the target is a group. Here's my thinking for why that's appropriate:

* Technically the AccountAssignment `principalId` field can accept the id of either an identitystore Group or an identitystore User.
* Assigning permissions to groups and not users is a near-universal best practice
* Because the identitystore Groups use a provider-generated UUID, it is very difficult to patch in their values without a selector, and a selector provides a very easy experience. In fact, providing the ability to use such a selector in my organization was one of the primary reasons I picked up this piece of work in the first place.
* Users who want to assign a permission set in a particular account to a user and not a group can still do so, just not using the selector, and their experience is no worse than if the selector didn't exist.

Maybe `PrincipalGroupId` would be better? I'm also fine with just leaving them as the default generated names if that's what you'd prefer.

config/identitystore/config.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"github.com/crossplane/upjet/pkg/config"
+)
+
+// Configure adds configurations for the identitystore group.
+func Configure(p *config.Provider) {
+	p.AddResourceConfigurator("aws_identitystore_group", func(r *config.Resource) {
+		// Display name is required by terraform, and while it's not part of the external name or terraform id, it is
+		// how the group is displayed, and it's immutable.
+		r.ExternalName.IdentifierFields = append(r.ExternalName.IdentifierFields, "display_name")
```
mbbush, line 12: Is there an easy, idiomatic golang way to append this only if it's not already present? In e.g. python I'd probably do it by converting to a set and then back to a list. Or maybe it's not worth worrying about since this happens at generation time.

Or maybe it's better to leave this off? I added it in an effort to get the field marked as required in the generated schema, but subsequently I realized that many fields which are actually required by terraform are not marked as required in the schema, but are instead validated at the kubernetes level using slightly more sophisticated logic that takes into account `initProvider` and `managementPolicies`. 

Conceptually it makes sense that the display name would be an identifier field. But it's not necessary to mark it as such in order for the managed resource to work properly.

config/ssoadmin/config.go
mbbush: This is extracting the `spec.forProvider.instanceArn` from the `PermissionSet` to use in the `PermissionSetInlinePolicy`. It does require repeating the same selector/reference twice in the managed resource manifest to do so, which is a bit awkward, and it's not obvious from the name that the `InstanceArnSelector` should actually point to a `PermissionSet`. It still seemed better than requiring the user to manually specify the instance arn an additional time. 

I didn't come up with this idea myself, actually. I noticed that upjet generated a reference identical to this one on one of the other resources, from the terraform example. 

Because PermissionSet is the one resource that is referenced by every other resource in this group, it made sense to me that we could use that as a source. Ultimately the "right" object to reference would be a terraform [data.aws_ssoadmin_instances](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ssoadmin_instances) resource, or perhaps a not-yet-existing resource version of that data source. 

I now notice that it's present on some but not all of the SSO resources, so I think it would be better to either add it where it's missing or take it off everywhere (including the places where it was automatically generated). I'm happy to do whichever you prefer.

examples/ssoadmin/customermanagedpolicyattachment.yaml
mbbush: ok, I'll add annotations to all the resources which depend on resources with manual intervention annotations. I think that will amount to all of the resources in all the examples except the IAM policies.

examples/ssoadmin/permissionsetinlinepolicy.yaml
mbbush: I'm not sure what you mean, but maybe my answer at https://github.com/upbound/provider-aws/pull/928#discussion_r1370683283 clarifies?

examples/ssoadmin/managedpolicyattachment.yaml
mbbush: I just realized that these examples aren't only used for uptest, they're also shown on the provider docs webpage, where the terraform syntax (which won't work) will be misleading. 

I'll make them all consistent and remove the tf syntax. If someone needs to reference it it's still in examples-generated.

mbbush:
@turkenf This is ready for another look.

config/ssoadmin/config.go
mbbush: I ended up taking it off.

config/ssoadmin/config.go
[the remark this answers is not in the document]
turkenf: The method we follow when defining a ref and selector for a parameter is usually as follows:
- If that parameter points exactly to a certain source, the reference and selector can be defined.
- If it points to more than one resource rather than a specific resource, we generally do not prefer to define a reference and selector. But if there is a common use for it, we can choose to define a reference and selector to the commonly used resource by evaluating user experiences and feedback.

The situation here means that we define a reference and selector according to the second case. In your usage, using `PrincipalGroup` for the reference and selector name is quite logical and understandable because this is only possible when we use the group ID. However, when I think of it as a user who is not familiar with the content and wants to simply create this resource, I think it would be confusing if the name of a parameter is different from the name of the reference and selector. That's why I prefer default names.

If we want to indicate to the user that the reference and selector are defined for the group, maybe we can add a configuration as follows.

```
		r.MetaResource.ArgumentDocs["principal_id"] = "- (Required) An identifier for an object in SSO, such as a user or group(The reference and selector are defined for the group, not for the user)."

```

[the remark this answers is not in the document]
turkenf: If we continue with the group, we can also give it a name like `PrincipalIdFromGroup` to make it understandable for the user, I leave the choice up to you.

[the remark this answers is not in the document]
turkenf: I'm not sure if there is a way to edit the document for the reference and selector, we can continue by editing the primary field for now.

mbbush:
@turkenf Ready for your review again. Is there anything else this needs before merging? I'm hoping to get this in the October release so I can start using these resources next week at work.

turkenf:
I am closing and reopening PR to run license/cla again.

package/crds/ssoadmin.aws.upbound.io_permissionsets.yaml
```diff
@@ -95,6 +95,7 @@ spec:
                     description: Key-value map of resource tags.
                     type: object
                 required:
+                - instanceArn
```
mbbush, line 98: This is a breaking change. It's a consequence of normalizing the external name for this resource.

From a practical point of view, this will only affect Observe only resources, because this parameter was always required by terraform, and therefore included in the kubernetes validation. If I make it no longer required, if this is missing, and the external name annotation is set to the permission set id (the normalized external name I chose), crossplane will be unable to construct the terraform id, which contains the instance arn, so it won't be able to import the existing resource.

I could remove it from the `IdentifierFields` in the external name configuration, which would avoid the breaking schema change, but it feels like it really _should_ be required, especially since omitting it will cause imports to fail. 

Another option would be to change the external name to include the instance arn (perhaps by changing it from the short id of the permission set to the full ARN of the permission set, which embeds the id of the instance, from which the instance arn can be derived). But that seems like sacrificing the long-term user experience in order to avoid a minor breaking api schema change.

mbbush: With this removed from `IdentifierFields` in the external name config, the following examples still work:
* A resource that specifies all the fields in `spec.forProvider` with default management policies
* A resource that doesn't include this field in `spec.forProvider`, but does include it in `spec.initProvider`, with default management policies (this field is immutable, so there's really no reason someone would *want* to put it in `spec.initProvider`)
* A resource with only the `Observe` management policy, that specifies the external name annotation and populates this field to the correct value in `spec.forProvider`

This doesn't work:
* A resource with only the `Observe` management policy, that specifies the external name annotation, but doesn't populate this field. This passes the kubernetes validator, but fails at runtime because it can't construct the terraform id.

mbbush: I can think of a handful of ways forward here:

* Normalize the external name to just the provider-generated part of the ARN, and leave `instanceArn` as an identifier field. If this was a new resource, as opposed to an update to an existing one, this seems like it would be the clear winner. 
* Normalize the external name, but don't specify `instanceArn` as an identifier field. This will cause an error for Observe resources as mentioned above.
* Use the permission set ARN as the external name (as opposed to the current external name which is `${permission_set_arn},${instance_arn}`. This could be done without specifying `instanceArn` as an identifier field because the instance arn can be derived from the permission set arn. 
* Leave the current external name which matches the terraform id. This would certainly be the smallest change, but it's also a fairly cumbersome external name to have to use. 

One factor that seems relevant to deciding is what happens to existing managed resources during a provider upgrade, when the external name config changes. I'm hoping that as long as the resources were ready, crossplane would simply reconstruct the external name from the terraform id, using the new logic, which would provide the correct result, but I'm not sure quite how best to test this. Would it be sufficient to take a MR that has become ready and simply edit its external-name annotation to the old format? Would there be other edits I should do at the same time? Or is it necessary to do a "full" test, where I actually do a provider upgrade on my test cluster?

config/externalname.go
```diff
@@ -2460,17 +2460,39 @@ var ExternalNameConfigs = map[string]config.ExternalName{
...
 	// SSO Permission Sets can be imported using the arn and instance_arn separated by a comma (,)
 	// Example: arn:aws:sso:::permissionSet/ssoins-2938j0x8920sbj72/ps-80383020jr9302rk,arn:aws:sso:::instance/ssoins-2938j0x8920sbj72
-	// TODO: Normalize external_name while testing
-	"aws_ssoadmin_permission_set": config.IdentifierFromProvider,
+	// All the information needed to generate the terraform id is in the parameters, but requires more parsing than the template supports.
+	"aws_ssoadmin_permission_set": SsoPermissionSet(),
+	// SSO Managed Policy Attachments can be imported using the name, path, permission_set_arn, and instance_arn separated by a comma (,)
+	// Example: TestPolicy,/,arn:aws:sso:::permissionSet/ssoins-2938j0x8920sbj72/ps-80383020jr9302rk,arn:aws:sso:::instance/ssoins-2938j0x8920sbj72
+	// This can't really be normalized.
+	"aws_ssoadmin_customer_managed_policy_attachment": config.TemplatedStringAsIdentifier("", "{{  (index .parameters.customer_managed_policy_reference 0).name }},{{ (index .parameters.customer_managed_policy_reference 0).path }},{{ .parameters.permission_set_arn }},{{ .parameters.instance_arn }}"),
```
mbbush, line 2476: This results in the crossplane external name matching the terraform id, which is quite long, but it occurs to me that perhaps it doesn't need to? 

This is quite possibly a bad idea, but it's not immediately obvious to me why we couldn't just have the `external-name` match `metadata.name`, and construct the terraform id by ignoring the external name and applying the template to the parameters.

mbbush:
@turkenf Anything else I can do to get this merged?

turkenf:
@mbbush, there is nothing you can do right now, it is waiting for the team to review, it will be reviewed as soon as possible, thank you.

sergenyalcin:
@mbbush Thank you for this contribution. I reviewed the PR. Let me talk about two parts of this. The first part is adding/configuring new resources. I think this part is ready for merge. The second part is changing the external name configurations for two resources. I am concerned about the breaking API changes. We do not want to introduce any breaking API changes in minor versions.

It seems that for the `PermissionSet` resource, we are introducing a new required field and deleting the X-Rules. As you mentioned, this is a breaking change. Also, you are correct that this was a required field in behavior because of the X-Rules.

In the mid/long-term, we want to switch to XValidation-Rules where possible so that fields behave as required. So, ideally, we want to leave no required fields in the schemas and ensure that the relevant fields act as mandatory with XValidation-Rules.

For the `PermissionSetInlinePolicy`, do we change the external-name syntax? If this is the case, this will also be a breaking change.

In short, if we address these normalizations without breaking changes (API or external name syntax), I think the normalization part will be okay for the minor version. However, if this is not the case and if we need some breaking changes for normalization, I will suggest separating the PR into two parts. And merging the adding new resources part here (for six resources). Then, we may consider handling the remaining external name normalizations in another PR for the major version releases.

mbbush:
Thanks for the review @sergenyalcin!

I've rebased on the latest main, removed the normalization of `PermissionSet`, discarded all my codegen commits and regnerated. 

The external name for PermissionSetInlinePolicy is changing, from `arn:aws:sso:::permissionSet/ssoins-2938j0x8920sbj72/ps-80383020jr9302rk,arn:aws:sso:::instance/ssoins-2938j0x8920sbj72` to just `ps-80383020jr9302rk`, however I believe that this is not actually a breaking change. As is clear from the lack of a diff in the generated code, the schema doesn't change, as in both cases the same two parameters are set as identifier fields. An existing managed resource with the old external name won't lose its reference to the terraform id, since `GetIDFn` will still create the desired id (in both the new and old versions the only relevant inputs were in `spec.forProvider`). It will just have its external name updated to the new value the next time `GetExternalNameFn` runs. 

The only thing I can think of possibly breaking as a result would be a composition or gitops manifest that was constructing the external name for an `Observe` resource and specifying it in the old format. Are we saying that the format of external names (which we don't actually document anywhere but the source code) is a part of the public API that we're committed to not changing? externalname.go is full of `// TODO normalize when testing` comments, so I assumed it was considered unstable. 

If you still think it's important to not change the external name for `PermissionSetInlinePolicy`, I can pull that change out too.  I would really like to get this included in the December release. (I originally wanted to get it in the October release!)

I'll rerun the tests tomorrow and edit the top comment with results.

mbbush:
@sergenyalcin this is rebased, tested, and ready for review with the changes you requested.

PR #933, opened 2023-10-22
fix(iam-role): ignore managed_policy_arns in late init

### LINKED ISSUE #929: RolePolicyAttachment.iam is detaching policies every reconciliation
### What happened?
Every time a `RolePolicyAttachment.iam.aws.upbound.io` resource that has a `policyArnSelector` is reconciled, the policy is detached and re-attached to the role. This results in IAM permissions being temporarily incorrect, causing unexpected errors and downtime for some applications. 

### How can we reproduce it?
This happens for all of our resources using a selector similar to this:

```yaml
spec:
  forProvider:
    policyArnSelector:
      matchControllerRef: true
      matchLabels:
        service: foo
```

If we do a `kubectl get rolepolicyattachments -w` we see that any resources doing this become unready every reconciliation loop, and the policies disappear from the resource in the AWS console or CLI. This last up to 30 seconds before being restored. The `provider-aws-iam` pod with debug logs enabled show the creation event, but not whatever is causing it to become detached.

* Crossplane Version: 1.12.2
* Provider Version: 0.41.0
* Kubernetes Version: v1.24.16-eks-2d98532

haarchri:
description is here: https://github.com/upbound/provider-aws/issues/929#issuecomment-1774063901
skip managed_policy_arns for late init

Fixes #929

I have:

Manual: https://github.com/upbound/provider-aws/issues/929#issuecomment-1774063901
Uptest: https://github.com/upbound/provider-aws/actions/runs/6611409372

### DIFF
--- a/apis/iam/v1beta1/zz_generated_terraformed.go
+++ b/apis/iam/v1beta1/zz_generated_terraformed.go
@@ -847,6 +847,7 @@ func (tr *Role) LateInitialize(attrs []byte) (bool, error) {
 		return false, errors.Wrap(err, "failed to unmarshal Terraform state parameters for late-initialization")
 	}
 	opts := []resource.GenericLateInitializerOption{resource.WithZeroValueJSONOmitEmptyFilter(resource.CNameWildcard)}
+	opts = append(opts, resource.WithNameFilter("ManagedPolicyArns"))
 
 	li := resource.NewGenericLateInitializer(opts...)
 	return li.LateInitialize(&tr.Spec.ForProvider, params)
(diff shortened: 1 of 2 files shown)


haarchri:
@turkenf @jeanduplessis please take a look. I believe we require a fix release to address this issue, as fluctuating role policies can cause problems in a production environment.

turkenf:

turkenf (approved):
Thank you @haarchri, LGTM.

mbbush:
My main question is whether we should also skip late init on `inline_policy`

turkenf:
> My main question is whether we should also skip late init on inline_policy

In order not to block this PR, we can handle `inline_policy` in another PR if it causes problems.

blakebarnett:
Thanks for the quick fix!

PR #965, opened 2023-11-15
Remove obsolete region field from organizations resources

### LINKED ISSUE #953: Obsolete `region` param for organization resources
### What happened?

For the AWS Account resource creation it is required to pass a valid region here:
https://marketplace.upbound.io/providers/upbound/provider-aws-organizations/v0.43.1/resources/organizations.aws.upbound.io/Account/v1beta1#doc:spec-forProvider-region

The AWS Accounts are not associated with any region, so the `region` param should not be a part of the input.

The same refers to all organizations' resources:

- https://marketplace.upbound.io/providers/upbound/provider-aws-organizations/v0.43.1/resources/organizations.aws.upbound.io/DelegatedAdministrator/v1beta1
- https://marketplace.upbound.io/providers/upbound/provider-aws-organizations/v0.43.1/resources/organizations.aws.upbound.io/OrganizationalUnit/v1beta1
- https://marketplace.upbound.io/providers/upbound/provider-aws-organizations/v0.43.1/resources/organizations.aws.upbound.io/Organization/v1beta1
- https://marketplace.upbound.io/providers/upbound/provider-aws-organizations/v0.43.1/resources/organizations.aws.upbound.io/Policy/v1beta1
- https://marketplace.upbound.io/providers/upbound/provider-aws-organizations/v0.43.1/resources/organizations.aws.upbound.io/PolicyAttachment/v1beta1

* Crossplane Version: -
* Provider Version: v0.43.1
* Kubernetes Version:  -
* Kubernetes Distribution:  -

lewyg:

Remove `region` field from input in organizations group resources. This field is automatically added for all resources, but for the organizational ones, should be skipped (similar to `iam` resources) because these resources are not associated with any region.

Currently, this is a required field and some value must be provided, even though it is not used by the resource.

Fixes #953

**This is a breaking change** because it removes the required parameter from org resources. Old compositions and examples will fail on schema validation with the error (example for account):
```
Account in version "v1beta1" cannot be handled as a Account: strict decoding error: unknown field "spec.forProvider.region"
```

I have:

Account creation test:

1. Created Account with old version (with region)
```yaml
apiVersion: organizations.aws.upbound.io/v1beta1
kind: Account
metadata:
  name: test-account
spec:
  forProvider:
    email: test-account@my-domain.com
    name: test-account
    region: us-west-1
```
2. Updated provider-aws (removed region from schema)
3. `region` has been automatically removed from the existing Account MR

After updating the provider, the above example is not working anymore, because of schema validation


ytsarev (changes requested):
@lewyg thanks a lot for the contribution! Can you please reflect the changes in the main[ examples](https://github.com/upbound/provider-aws/tree/main/examples/organizations) directory? These are the ones we use for uptest and they will also be packaged as examples for eventual Marketplace exposure

jbw976 (commented):
Does this make the `region` field no longer required, or does it remove it from the schema entirely? I'm not super familiar with all the upjet machinery 😇 

I ask because I'm curious how existing Organizations resources/manifests will behave after this change if they are still setting the `region` field. Will that be accepted, so this isn't a breaking change? 🤔

lewyg:
> Does this make the `region` field no longer required, or does it remove it from the schema entirely? I'm not super familiar with all the upjet machinery 😇
> 
> I ask because I'm curious how existing Organizations resources/manifests will behave after this change if they are still setting the `region` field. Will that be accepted, so this isn't a breaking change? 🤔

@jbw976 this removes `region` from the schema entirely, so I think it is a breaking change since it requires removing region field from existing manifests/compositions. 
For the existing resources, based on my local tests, `region` has been removed automatically.

lewyg:
> @lewyg thanks a lot for the contribution! Can you please reflect the changes in the main[ examples](https://github.com/upbound/provider-aws/tree/main/examples/organizations) directory? These are the ones we use for uptest and they will also be packaged as examples for eventual Marketplace exposure

done, thanks for pointing it out ;)

ytsarev (approved):
Thanks a lot! All of the example resources have `manual-intervention` annotation and I trust the e2e test performed by @lewyg for these resources.

turkenf:
Many thanks for your effort in this PR @lewyg, It would be nice if you could fill in the 'How has this code been tested' section in the description about how you tested it.

When we try to run the old examples with this PR, we will receive the following error:
```
Error from server (BadRequest): error when creating "/var/folders/sl/d7p3l_010ld35kpzf77017740000gn/T/tmp.MaOxyZWUm7": Account in version "v1beta1" cannot be handled as a Account: strict decoding error: unknown field "spec.forProvider.region"
```

The important thing here is that a previously working composition containing one of these resources will not work without removing the region field.

jeanduplessis:
We need to hold off on introduce this change.

We are about to release a major architecture change in this provider and I don't want to conflate the new architecture release with breaking changes from an unrelated change.

We are also looking into how we can more gracefully deal with breaking changes by leveraging multi-version CRDs and conversion webhooks.

Following the upcoming release with the new architecture we are targeting an upgrade to the latest Terraform AWS v5 provider which itself will also bring many breaking changes. We should consider incorporating this change as part of that effort.

lewyg:
@turkenf Yes, it will not work with older examples. I will update the descriptions with more details.

@jeanduplessis Since this is a breaking change, I agree that it's a good idea to wait until the release of the latest terraform AWS provider.

In the meantime, I discovered that this problem also occurs with other non-regional AWS services, for example: Route53 (would be good to go through other services and check them):
 - https://marketplace.upbound.io/providers/upbound/provider-aws-route53/v0.43.1/resources/route53.aws.upbound.io/Zone/v1beta1#doc:spec-forProvider-region
 -  https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route53_zone

What do you think about making it optional for now, instead of removing it? I will not break anything, but will enable me to drop `region` when not needed?

jeanduplessis:
@lewyg I think it's reasonable to make it optional as it will still be backwards compatible.

lewyg:
Ok, then I'm closing this PR, because the change requires better preparation

PR #979, opened 2023-11-21
Bump golangci-lint to v1.55.2

ulucinar:

Fixes #

I have:

It looks like bumping the `golangci-lint` version to the latest did not resolve the OOM issue. But we still decided to bump our dependency and we will probably do a memory profiling to understand what's causing the memory surge in the linter.


turkenf (approved):
Thank you @ulucinar, LGTM.

Issue #997, opened 2023-12-01
Excessive calls to AssumeRoleWithWebIdentity w/ IRSA


haarchri:
### What happened?
Reduce the frequency of sts calls for AssumeRole/AssumeRoleWithWebIdentity by utilizing cached credentials. This will minimize the number of audit events in CloudTrail, and only renew the credentials when necessary, such as before expiration.

check implementation in community provider https://github.com/crossplane-contrib/provider-aws/issues/828

### What environment did it happen in?

* Crossplane Version:
* Provider Version:
* Kubernetes Version:  
* Kubernetes Distribution:

haarchri:
@erhancagirici we have no cache enabled - we want to reopen this issue ? or create a new ?

stevendborrelli:
The fix in #1003 was a partial fix. We are still seeing significant calls to `AssumeRoleWithWebIdentity`, which propagates across audit and logging systems.

haarchri:
good to check https://github.com/crossplane-contrib/provider-aws/pull/1148

erhancagirici:
@haarchri As a first step, the refactoring at https://github.com/upbound/provider-aws/pull/1204 will reduce the STS calls by 50%.

I am working on a cache implementation and I've checked the https://github.com/crossplane-contrib/provider-aws/pull/1148. I've been using a similar approach but there seems to be several things to consider with that and ongoing work, especially for some edge cases:

Community provider implementation relies on caching the constructed default AWS config.
- this only covers plain IRSA. Other options like IRSA+AssumeRole, WebIdentity, WebIdentity+AssumeRole configurations still do not benefit from the cache.
- Default AWS config is unconditionally cached and there is no concept of invalidation of that cached config. While it holds for most of the time (for IRSA), this has a big assumption that the configuration never changes on runtime.

Some background info (on the implementation details):
AWS config struct has the field `Credentials` which is an `CredentialProvider` interface that implements `Retrieve()` method.
By default, when AWS SDK is constructing an AWS Config (e.g. via `LoadDefaultConfig` ) it wraps the resulting `CredentialProvider` with a simple `CredentialsCache`, which is a credential provider itself. That cache is only for the wrapped CredentialProvider, for the particular credential config.

When we cache the config itself on top of that, we end up with a layered cache, i.e. we cache the underlying `CredentialCache`, and the wrapped `CredentialProvider` as a side effect.

From this point on speaking for the all auth cases in a generic manner:
[...]

PR #1003, opened 2023-12-04
refactor external client to make single getAWSConfig call per connect 

erhancagirici:

Refactors `getAWSConfig` calls in the external client code, to make single call per connect. `getAWSconfig` calls involve resolving credentials and executing AWS STS calls when there are relevant `AssumeRole` or `AssumeRoleWithWebIdentity`  operations needed. This was being called multiple times per connect, causing superfluous invocations.

This PR will improve the situation reported at #997

I have:

Via uptest: https://github.com/upbound/provider-aws/actions/runs/7141372941

### DIFF
--- a/internal/clients/aws.go
+++ b/internal/clients/aws.go
@@ -33,10 +33,6 @@ const (
 	keyRoleArn                   = "role_arn"
 	keySessionName               = "session_name"
 	keyWebIdentityTokenFile      = "web_identity_token_file"
-	keyAssumeRole                = "assume_role"
-	keyTags                      = "tags"
-	keyTransitiveTagKeys         = "transitive_tag_keys"
-	keyExternalID                = "external_id"
 	keySkipCredsValidation       = "skip_credentials_validation"
 	keyS3UsePathStyle            = "s3_use_path_style"
 	keySkipMetadataApiCheck      = "skip_metadata_api_check"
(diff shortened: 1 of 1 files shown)


haarchri:
Is it possible to implement credential caching? Currently, it seems we make a single call per connect. However, when managing multiple resources, each resource triggers a separate call. Considering that IRSA credentials are valid for 15 minutes, it would be more efficient to perform the call only once every 15 minutes, even when dealing with numerous managed resources - what do you think ?

erhancagirici:
@haarchri This PR is an initial improvement for addressing the low-hanging fruit. Our plan is to make further improvements in iterations. This is especially motivated by the fact that those parts of the provider are currently fractured because we have a hybrid architecture for that provider (both the CLI-based & SDK-based upjet architectures are in use at the moment). 
The plan is once we completely switch to the nofork arch for the remaining few resources , we will be able to simplify the provider setup logic there. And as we simplify provider setup logic, it will be easier for us to make further optimizations like credential caching, that is definitely in our scope.

erhancagirici:

internal/clients/aws.go
```diff
@@ -199,7 +196,7 @@ func DefaultTerraformSetupBuilder(ctx context.Context, c client.Client, mg resou
 	if pc.Spec.Endpoint != nil {
 		if pc.Spec.Endpoint.URL.Static != nil {
 			if len(pc.Spec.Endpoint.Services) > 0 && *pc.Spec.Endpoint.URL.Static == "" {
-				return errors.Wrap(err, "endpoint is wrong")
```
ulucinar, line 202: Good catch! Then, we were previously silently non-erroring :)

ulucinar (approved):
Thanks @erhancagirici, lgtm.

PR #1008, opened 2023-12-05
feat(ref): add kms ref/selector in replication configuration

haarchri:
add kms ref/selector in replication configuration for `aws_s3_bucket_replication_configuration`

Fixes #

I have:


config/s3/config.go
```diff
@@ -104,6 +104,9 @@ func Configure(p *config.Provider) {
 			Type:      "github.com/upbound/provider-aws/apis/s3/v1beta1.Bucket",
 			Extractor: `github.com/crossplane/upjet/pkg/resource.ExtractParamPath("arn",true)`,
 		}
+		r.References["rule.destination.encryption_configuration.replica_kms_key_id"] = config.Reference{
+			Type: "github.com/upbound/provider-aws/apis/kms/v1beta1.Key",
```
sergenyalcin, line 108: If I understand correctly, we need to capture the ID of kms.Key resource as the reference value. Without overriding the extractor, we will use the default `ExternalName` extractor. What do you think about adding the ID extractor: 

```suggestion
			Type: "github.com/upbound/provider-aws/apis/kms/v1beta1.Key",
			Extractor: common.PathTerraformIDExtractor,
```

sergenyalcin (commented):
Thanks @haarchri I left a comment

config/s3/config.go
haarchri: in the field description for replica_kms_key_id i can see 

`	// ID (Key ARN or Alias ARN) of the customer managed AWS KMS key stored in AWS Key Management Service (KMS) for the destination bucket.`

haarchri:
@sergenyalcin i can rebase - can we merge this after ?

config/s3/config.go
sergenyalcin: As I see, the external name of kms.Key is arn of this resource so the default extractor works.

sergenyalcin (approved):
Thanks @haarchri LGTM! After rebasing we can merge this.

turkenf:

PR #1015, opened 2023-12-07
Add iot TopicRuleDestination to v1beta1

### LINKED ISSUE #361: Moving iot (1) resource to v1beta1 version
External name configuration of 1 resource in the `iot` group:
- [ ] [`aws_iot_topic_rule_destination`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iot_topic_rule_destination)

mbbush:

Fixes #361 

I have:

- [x] Run `make check-diff lint test` to ensure this PR is ready for review.

`make e2e UPTEST_EXAMPLE_LIST=examples/iot/topicruledestination.yaml`

The test takes about 15 minutes to run, due to AWS being slow to create and delete these resources. The equivalent acceptance test in the terraform provider takes basically the same amount of time. 

Note that in the example, I used a management policy that excludes `Delete` for the IAM role. This is necessary to allow the IoT service to assume the role during deletion of the topic rule destination, to delete the ENIs it created. Otherwise deletion of the VPC fails because there are leftover resources in it. The result is that the IAM role won't be cleaned up. But that shouldn't be a problem. It's got a name that's unlikely to conflict with anything else, and rerunning the test a second time when the role already exist works just fine. That's one reason I used an inline policy, so I could only have a single leftover resource, rather than also leaving behind an IAM policy.


turkenf:
@mbbush, since you are now a contributor, you should be able to trigger uptest on open PRs.

mbbush:
/test-examples="examples/iot/topicruledestination.yaml"

mbbush:
/test-examples="examples/iot/topicruledestination.yaml"

turkenf (commented):
Thank you @mbbush, I left a few comments for you to consider.

config/externalname.go
```diff
@@ -2081,6 +2078,9 @@ var NoForkExternalNameConfigs = map[string]config.ExternalName{
 	"aws_iot_thing_type": config.IdentifierFromProvider,
 	// IoT Topic Rules can be imported using the name
 	"aws_iot_topic_rule": config.NameAsIdentifier,
+	// IoT topic rule destinations can be imported using the arn
+	// arn:aws:iot:us-west-2:123456789012:ruledestination/vpc/2ce781c8-68a6-4c52-9c62-63fe489ecc60
+	"aws_iot_topic_rule_destination": TemplatedStringAsProviderDefinedIdentifier("arn:aws:iot:{{ .setup.configuration.region }}:{{ .setup.client_metadata.account_id }}:ruledestination/vpc/{{ .external_name }}"),
```
mbbush, line 2083: What do you think about the relative merits of using just the provider-generated UUID suffix as the external name (as I've done here) vs the full arn (which is also the terraform id)? The arn is certainly longer, and more cumbersome, but also perhaps easier for new users to discover, since we don't currently have any way of documenting the expected format for an external name.

examples/iot/topicruledestination.yaml
```diff
@@ -0,0 +1,165 @@
...
+  labels:
+    testing.upbound.io/example-name: iot-topic-rule-destination
+  name: iot-topic-rule-destination-example
+spec:
+  managementPolicies:
+    - Create
+    - Update
+    - Observe
+    - LateInitialize
+    # It is necessary to orphan-delete the role so that the TopicRuleDestination has permission to clean up after itself during deletion.
```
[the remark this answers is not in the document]
mbbush: That works well, thanks for the suggestion.

mbbush:
/test-examples="examples/iot/topicruledestination.yaml"

mbbush:
@turkenf Thanks for the testhook recommendation. It worked great. 

I think the only thing left before this can merge is a decision on whether to use the short identifier or the longer but more "discoverable" full ARN as the external name. I see pros and cons with both options, and am willing to go either way.

config/externalname.go
turkenf: I am okay with the above configuration and we can continue with this.

turkenf (approved):
Thank you @mbbush, LGTM.

PR #1070, opened 2024-01-06
DocDB: add `dbClusterParameterGroupName` Selector for Cluster

ytsarev:

Add `dbClusterParameterGroupNameSelector` to DocDB Cluster

* Add reference configuration
* Add generated files
* Extend uptest example

I have:

uptest below

### DIFF
--- a/config/docdb/config.go
+++ b/config/docdb/config.go
@@ -17,6 +17,9 @@ func Configure(p *config.Provider) {
 			}
 			return conn, nil
 		}
+		r.References["db_cluster_parameter_group_name"] = config.Reference{
+			TerraformName: "aws_docdb_cluster_parameter_group",
+		}
 	})
 
 	p.AddResourceConfigurator("aws_docdb_cluster_instance", func(r *config.Resource) {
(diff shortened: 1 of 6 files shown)


config/docdb/config.go
```diff
@@ -17,6 +17,9 @@ func Configure(p *config.Provider) {
 			}
 			return conn, nil
 		}
+		r.References["db_cluster_parameter_group_name"] = config.Reference{
+			Type: "DBClusterParameterGroup",
```
haarchri, line 21: nit: When defining the cross-resource references, we have a more "robust" way of specifying the reference target type. You may also use the config.Reference.TerraformName instead of config.Reference.Type. We consider it as more robust because the generated type names/groups may change via upjet's resource configuration API, whereas the Terraform name is more stable.

ytsarev: Thanks, will update the reference.

ytsarev:
/test-examples="examples/docdb/cluster.yaml"

ytsarev:
/test-examples="examples/docdb/cluster.yaml"

ulucinar (approved):
Thanks @ytsarev, lgtm.

haarchri (approved):
Thanks - LGTM

config/docdb/config.go
mbbush: Thanks for the explanation! I've noticed this preference in some PR comments, but never understood why this method was preferred.

PR #1021, opened 2023-12-13
(bugfix) Fix cognito user pool client orphaned resources

mbbush:
### Description of your changes

The cognito idp user pool client is one of the handful of resources in terraform-provider-aws v4.67.0 that use the terraform plugin framework, and therefore that we still reconcile by forking a CLI terraform process. 

There are three problems with the cognito user pool client:
1. Creation fails if the `spec.forProvider.name` doesn't match the regex for the `clientId`
2. Whenever the provider pod restarts, crossplane will either create a duplicate resource, orphaning the existing one, or get into a Synced: False state, depending on whether `spec.forProvider.name` matches the `clientId` regex. 
3. Observe-only resources fail, because `terraform import` expects an argument that is not the same as the terraform id. 

This PR fixes the first two issues, and makes the third issue no worse. 

### Context

This resource has two idiosyncrasies that make it difficult. 

1. Calling `terraform apply -refresh-only` with an id of `""` returns an error, because the terraform provider passes through the empty-string id to the aws sdk, which (correctly) complains that trying to find a user pool client with id `""` is invalid. I suspect this is *because* the resource uses the terraform plugin framework, which introduced the concept of a null value as distinct from a zero value. I suspect that if we had some way to specify a null value for the id, it would work fine, but the type in upjet is `string` not `*string` so I don't see how to do that without major changes. Maybe the upjet developers will need to add a transformation layer from zero values back to null when integrating with the terraform provider framework, or perhaps they'll come up with a better idea. 

2. The argument needed to `terraform import` the resource is different than the terraform id, but the terraform id is needed to run `terraform apply -refresh-only`. The provider uses `terraform import` for resources with management policy `Observe` and `terraform apply -refresh-only` for other management policies. Unless someone has a clever way to identify which case we're in, in the context of an external name configuration, we can only support one of these use cases.

#762 was an attempt to work around the `"id":""` issue, but it introduced new problems.

My solution for the first issue is to simply use a static client id of "invalidnonemptystring", which satisfies the client id regex `[\w+]+`, so doesn't cause an error when we run `terraform apply -refresh-only`, but also doesn't match any actual existing client id (the aws-generated ids that I've seen are all 26 lowercase alphanumeric characters long). This feels like getting the right result by doing the wrong thing, but I can't think of any actual problems it would cause, nor can I come up with a better idea. 

[...]

PR #1072, opened 2024-01-09
complete example for securitygrouprule with self=null

mbbush:

Addresses the minor PR comments left on #1021, and adds a missing example useful for demonstrating #1073
 
I have:


examples/cognitoidp/userpoolclient.yaml
```diff
@@ -1,28 +1,45 @@
...
 metadata:
   labels:
-    testing.upbound.io/example-name: example
-  name: example
+    testing.upbound.io/example-name: userpoolclient
+  name: userpoolclient-with-dashes
 spec:
   forProvider:
-    name: example
+    name: name-that-doesnt-match-id-regex
```
mbbush, line 40: @turkenf I thought it worthwhile to have examples with names that both passed and failed the id regex, as they each displayed different problematic behavior before my bugfix (orphaned resources vs sync errors). Now that bug is fixed, but it makes sense to me to leave this around as a regression test.

I thought just combining both in the same file would work well, but then I'm not quite sure what to do about the example-id annotation. As I understand it, Uptest looks for a match between that annotation and the gvk of the resource to determine "is this the root resource of the example". Since this example has two resources with the same GVK as the "root", I thought maybe I should leave the example id off the second one? What do you think?

I'm happy to do whatever you prefer here.

turkenf: The purpose of using the example-id annotation was to show resources with their dependencies in the marketplace. But even if this approach is not implemented now, we want our samples to be ready if we continue to show resources with their dependencies on the marketplace in the future.

I think the example containing dash/dashes will be sufficient here as it covers both cases, so we can continue with an example containing dash/dashes in its name, what do you think?

mbbush: > I think the example containing dash/dashes will be sufficient here as it covers both cases

I don't understand what you mean. Before [the bugfix](https://github.com/upbound/provider-aws/pull/1021), security group rules with a `spec.forProvider.name` that matched the id regex created a new user pool client each time the crossplane provider pod restarted, while rules with a name that didn't match the regex refused to create user pool clients at all.

turkenf: > I don't understand what you mean. Before https://github.com/upbound/provider-aws/pull/1021, security group rules with a spec.forProvider.name that

I think you mean `UserPoolClient` above, so let me ask, does the `spec.forProvider.name: userpoolclient` example cover a situation that the `spec.forProvider.name: name-that-doesnt-match-id-regex` example does not? If there is no special error situation covered by the example `spec.forProvider.name: userpoolclient`, the example `spec.forProvider.name: name-that-doesnt-match-id-regex` will be sufficient. If not, let's add the two examples separately as in the previous PR.

mbbush: :facepalm: Yes, by security group rule I mean user pool client. Thanks for knowing what I mean even when it's not what I say :wink:.

Yes, there is a category of bugs related to the external name of the UserPoolClient which will be triggered by `spec.forProvider.name: userpoolclient` and which will not be triggered by `spec.forProvider.name: name-that-doesnt-match-id-regex`. I _think_ that any regression that would cause one to fail would also cause the other to fail (although in a different way), but I'm not certain of this and would rather just leave them both in, especially as this is a quick and cheap resource to create/destroy, unlike say a database, or a kafka cluster, or even an ec2 instance.

mbbush:
/test-examples="examples/cognitoidp/v1beta1/userpoolclient.yaml,examples/ec2/v1beta1/securitygrouprule.yaml"

mbbush:
/test-examples="examples/cognitoidp/v1beta1/userpoolclient.yaml,examples/ec2/v1beta1/securitygrouprule.yaml"

mbbush:
@turkenf Tests pass. This is ready to merge.

turkenf (approved):
Thanks @mbbush, LGTM.

PR #1107, opened 2024-01-24
fix Secret/secretmanagers.aws update loop for replica field

### LINKED ISSUE #1079: AWS Secrets Manager Provider Constantly Adds and Deletes Regional Replication
### What happened?
On upgrade to 0.47.1 (also observed the same behaviour in 0.46.0), the AWS Secrets Manager provider constantly removes regional replication for secrets and then adds it back. This results in a steep increase in billing costs for the AWS Secrets Manager API. Tailing the logs for the provider will show the constant creation and deletion of replication. Downgrading the provider to 0.43.0 resolved the issue.

### How can we reproduce it?
Create a [Secret](https://marketplace.upbound.io/providers/upbound/provider-aws-secretsmanager/v0.47.1/resources/secretsmanager.aws.upbound.io/Secret/v1beta1) resource with the replica config block set to replicate the secret to an additional region, ex:

```
---
apiVersion: secretsmanager.aws.upbound.io/v1beta1
kind: Secret
metadata:
  name: my-secret
spec:
  deletionPolicy: Delete
  forProvider:
    name: my-secret
    region: us-east-1
    replica:
    - region: us-east-2
```

* Crossplane Version: 0.14.5
* Provider Version: 0.47.1 (also observed for 0.46.0)
* Kubernetes Version: 1.28

erhancagirici:

Configures a custom diff function for `replica` field in `Secret/secretmanagers.aws` resource, with a special set difference implementation.

The `replica` attribute has `Set` type in the resource schema, which consists of elements with `region` and `kms_key_id` attributes. The `kms_key_id` is an optional + computed field, meaning that user can set a kms_key_id or if they don't, the cloud API will assign an automatic kms_key_id.

Terraform SDK has a limitation that set elements with optional+computed fields always show perma-diff, because of the Set representation and hashing logic. 
- Set diffs are calculated as a whole, by comparing the hash list of elements. Therefore, it never captures individual element field change diffs. As a result of that, changes in sets always result in removal of the changed element and recreated with the changed spec.
- Elements have their hash functions derived from their specifiable fields.

As a result of above, when optional+computed fields is not specified at resource config (spec.forProvider), they are considered empty at diffing/planning time, because there is no way of knowing that since it is computed by the cloud API. This causes their hash to be different from the one in the state. Therefore, the diff shows removal of the current one, and creation of a new one.

The implemented custom diff function, tries to match the current state replicas with the ones that do not have `kms_key_id` specified (after filtering the ones with both `region` and `kms_key_id` specified). 

Fixes #1079

I have:

Tested with the provided MR spec at issue #1079 and no update loop was detected

### DIFF
--- a/config/secretsmanager/config.go
+++ b/config/secretsmanager/config.go
@@ -1,9 +1,18 @@
 package secretsmanager
 
-import "github.com/crossplane/upjet/pkg/config"
+import (
+	"fmt"
+	"strconv"
+
+	"github.com/crossplane/upjet/pkg/config"
+	"github.com/crossplane/upjet/pkg/resource/json"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/pkg/errors"
+)
 
 // Configure adds configurations for the secretsmanager group.
-func Configure(p *config.Provider) {
+func Configure(p *config.Provider) { //nolint:gocyclo
 	p.AddResourceConfigurator("aws_secretsmanager_secret", func(r *config.Resource) {
 		// Use aws_secretsmanager_secret_rotation.
 		config.MoveToStatus(r.TerraformResource, "rotation_rules", "rotation_lambda_arn")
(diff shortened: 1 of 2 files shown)


config/secretsmanager/config.go
```diff
@@ -13,5 +22,193 @@ func Configure(p *config.Provider) {
 		r.TerraformConfigurationInjector = func(_ map[string]any, params map[string]any) {
 			params["name_prefix"] = ""
 		}
+		r.TerraformCustomDiff = func(diff *terraform.InstanceDiff, state *terraform.InstanceState, config *terraform.ResourceConfig) (*terraform.InstanceDiff, error) {
+			// skip diff customization on create
+			if state == nil || state.Empty() {
+				return diff, nil
+			}
+			if config == nil {
+				return nil, errors.New("resource config cannot be nil")
```
mbbush, line 31: If not nil, what is the `config` during a destroy call?

config/secretsmanager/config.go
```diff
@@ -13,5 +22,193 @@ func Configure(p *config.Provider) {
...
+				Region   string `json:"region"`
+			}
+
+			// traverse state
+			for _, v := range currentReplicaSet.List() {
+				// v is an interface{} type, which is a replica
+				// marshal then unmarshal to convert it to the internal type replica for easier access to fields
+				replicaBytes, err := json.JSParser.Marshal(v)
+				if err != nil {
+					continue
```
mbbush, line 71: Shouldn't this return the serialization error?

config/secretsmanager/config.go
```diff
@@ -13,5 +22,193 @@ func Configure(p *config.Provider) {
...
+				}
+				cReplica := &replica{}
+				if err := json.JSParser.Unmarshal(replicaBytes, cReplica); err != nil {
+					return nil, err
+				}
+				if cReplica.Region == "" || cReplica.KMSKeyID == "" {
+					// we should not be here, replicas at current state always have their region set and kms_key_id computed
+					continue
```
mbbush, line 79: It seems like we should do _something_ if we somehow get here, at least logging a warning.

config/secretsmanager/config.go
```diff
@@ -13,5 +22,193 @@ func Configure(p *config.Provider) {
...
+				if dReplica.KMSKeyID == "" {
+					if count, ok := desiredReplicasWithRegionOnly[dReplica.Region]; ok {
+						desiredReplicasWithRegionOnly[dReplica.Region] = count + 1
+					} else {
+						desiredReplicasWithRegionOnly[dReplica.Region] = 1
+					}
+				} else {
+					// this is a Replica at params, with explicit KMS Key ID specified in region
+					// check whether we have an exact match in current state
+					if region, ok := regionsInCurrentState[dReplica.Region]; ok {
```
mbbush, line 126: It took me a bit to figure out that `region` didn't actually have a region in it, it was just a map from kms key id -> hash. I think a different name would be better, maybe `keyHashes` or something?

config/secretsmanager/config.go
```diff
@@ -13,5 +22,193 @@ func Configure(p *config.Provider) {
...
+					}
+				default:
+					// for the particular region, we have matching number of KMS Key IDs to desired, i.e. there should be no diff for these
+					// example
+					// current state for region2 KMS Key IDs = { region2_kmsKeyX}
+					// desired state for region2 KMS Key IDs = { region2_kmsKeyANY }
+					// due to set difference implementation in TF, this will show up as 1 deletion, 1 creation in DIFF
+					// instead, in this case, we want to have no diff
+					// thus, remove all creation diffs and remove all deletion diffs
+					// choose Replica to delete randomly
```
mbbush, line 183: I don't think this comment applies here.

config/secretsmanager/config.go
```diff
@@ -13,5 +22,193 @@ func Configure(p *config.Provider) {
...
+
+				switch {
+				case len(KMSKeysOfRegionInCurrentState) > unmatchedDesiredKMSKeyCount:
+					// for the particular region, we have more KMS Key IDs present in the current state than we desire, e.g.
+					// current state for region1 = { region1_kmsKeyA, region1_kmsKeyB, region1_kmsKeyC }
+					// desired state for region1 = { region1_kmsKeyANY }
+					// due to set difference implementation in TF, this will show up as 3 deletions, 1 creation in DIFF
+					// instead, in this case, we want to have 2 deletions, 0 creation DIFF
+					// thus, remove all (unmatchedDesiredKMSKeyCount=1) creation diffs and remove (unmatchedDesiredKMSKeyCount=1) deletion diff
+					// choose Replica to delete randomly
```
mbbush, line 156: It seems like you're choosing deterministically by hash. I wouldn't call that random.

config/secretsmanager/config.go
```diff
@@ -13,5 +22,193 @@ func Configure(p *config.Provider) {
...
+						"region":     region,
+					})
+					removeReplicaFromDiffViaHash(creationHash)
+				}
+			}
+			// compare the total desired Replica count and current Replica count
+			// adjust the diff for replica.#
+			if len(desiredReplicaList) == len(currentReplicaSet.List()) {
+				// no diff, therefore remove diff if exists
+				delete(diff.Attributes, "replica.#")
```
mbbush, line 198: Is this still the correct action to take if we're going from e.g. 
`current: region1_kmsKeyX`
`desired: region2_kmsKeyAny`

mbbush (commented):
Thanks @erhancagirici, This looks entirely correct, and while it's not nearly as simple as I'd like it to be, I also don't see any way to simplify it, other than building it into upjet. I especially appreciated your comments, which were just the right level of verbosity to help me understand why you were doing things.

I've got two lingering questions, though. 
1. Why didn't this same issue happen with provider version 0.43, when we were forking a terraform CLI?
2. How many other resources suffer from this same problem?

It seems like the answer to the second question is that this will happen for any resource with an element of type `schema.TypeSet`, whose elements contain at least one `Optional: true, Computed: true` field and at least one `Computed: false` field. It also seems like (assuming this works as intended) this approach could be moved into upjet to apply to any of those cases where there's exactly one `Optional: true, Computed: true` field and exactly one `Computed: false` field, which is also `Required: true`. 

I can take a search through the schema to see if there are other resources that may suffer from the same behavior, unless you've already done that.

config/secretsmanager/config.go
erhancagirici: Though destroy does not make use of it, it is always populated with the current parameters at `spec.forProvider`
https://github.com/crossplane/upjet/blob/87fae5ea613cb8908e976c8f2a8c966e3416f96e/pkg/controller/external_nofork.go#L418
and passed to `TerraformCustomDiff` function
https://github.com/crossplane/upjet/blob/87fae5ea613cb8908e976c8f2a8c966e3416f96e/pkg/controller/external_nofork.go#L424
For destroys, the indicator is the `Destroy` field of instanceDiff object
https://github.com/crossplane/upjet/blob/87fae5ea613cb8908e976c8f2a8c966e3416f96e/pkg/controller/external_nofork.go#L689

config/secretsmanager/config.go
erhancagirici: agreed, a bit verbose but maybe something like `KMSKeysWithElemHashesOfRegion`, because `keyHashes` might make one think of a hash of the key itself.

config/secretsmanager/config.go
erhancagirici: yes, maybe the word "arbitrarily by the map iteration order" would be technically more correct. My intention was to mention that from user's pov the extra replicas to delete would be kind of "deterministically random 😅" in practice since the kms key IDs are afaik random and their addition to map influences iteration order and no special logic is applied for choosing what to delete.

config/secretsmanager/config.go
erhancagirici: yes, because there is no diff in in the set element count between old and new set. Also I tested this scenario.

FWIW, while working on other diffs in the past, we observed that this `setname.#` does have any effect in practice and ends up with correct elements.

config/secretsmanager/config.go
```diff
@@ -13,5 +22,193 @@ func Configure(p *config.Provider) {
...
+					// thus, remove all creation diffs and remove all deletion diffs
+					// choose Replica to delete randomly
+					for _, hash := range KMSKeysOfRegionInCurrentState {
+						removeReplicaFromDiffViaHash(hash)
+					}
+					creationHash := hashFunc(map[string]interface{}{
+						"kms_key_id": "",
+						"region":     region,
+					})
+					removeReplicaFromDiffViaHash(creationHash)
```
erhancagirici, line 191: theoratically, in the generic logic, this should remove all creation diffs, however practically this is never > 1. realized that we anyways cannot specify two replicas in the same region without kms ids specified in the config due to they end up as same hash and a single element. So, something like below is not possible to specify.
```yaml
apiVersion: secretsmanager.aws.upbound.io/v1beta1
kind: Secret
metadata:
  name: my-secret
spec:
  deletionPolicy: Delete
  forProvider:
    name: my-secret
    region: us-east-1
    replica:
    - region: us-east-2
    - region: us-east-2
```
Anyways, this is not valid from the cloud API perspective, as you can have a single replica per region.

erhancagirici:
@mbbush thanks so much for the thorough review! much appreciated.

> Thanks @erhancagirici, This looks entirely correct, and while it's not nearly as simple as I'd like it to be, I also don't see any way to simplify it, other than building it into upjet. I especially appreciated your comments, which were just the right level of verbosity to help me understand why you were doing things.

glad that the comments helped through the review 🎉 , I intentionally left them at this verbosity, because although the high-level aim of the function is easy to word, the details are a bit cumbersome.

> 1. Why didn't this same issue happen with provider version 0.43, when we were forking a terraform CLI?

`aws_route53_resolver_endpoint` was also suffering with the same problem, and it was also there for CLI-based provider.
I think terraform relies on "not constantly reconciling" and they end up in plan anyways. Tried to find if there is any kind of special handling for this (maybe at the CLI level) but could not find that.

> 2. How many other resources suffer from this same problem?
 
> It seems like the answer to the second question is that this will happen for any resource with an element of type `schema.TypeSet`, whose elements contain at least one `Optional: true, Computed: true` field and at least one `Computed: false` field. 

Yes, there is also another factor: which fields does the specified hash function uses for that set. If the set hash function does not make use of the optional+computed field, the resource is not affected.

I've implemented a quick schema traverser and gathered a list. Need to go over the hash functions though.

> It also seems like (assuming this works as intended) this approach could be moved into upjet to apply to any of those cases where there's exactly one `Optional: true, Computed: true` field and exactly one `Computed: false` field, which is also `Required: true`.

I had a similar implementation done for `aws_route53_resolver_endpoint`, so the logic is borrowed from there. Making this as a part of upjet is also considered at our side.

config/secretsmanager/config.go
mbbush: One strategy I sometimes use for naming maps is `ValuesByKey`, so maybe something like `hashesByKMSKey`?

config/secretsmanager/config.go
mbbush: The most succinct and correct way I can think of to describe that would be 
```
// Arbitrarily choose which replica to delete, since they're indistinguishable
```

examples/secretsmanager/v1beta1/secret-withreplica.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: secretsmanager.aws.upbound.io/v1beta1
+kind: Secret
+metadata:
+  name: example
```
turkenf, line 4: nit: It would be better if we randomize the `metadata.name` field and add `example-id` annotation.

```suggestion
  name: example-${Rand.RFC1123Subdomain}
  annotations:
    meta.upbound.io/example-id: secretsmanager/v1beta1/secret
```

turkenf (approved):
Thank you @erhancagirici, LGTM.

sergenyalcin (approved):
Thanks @erhancagirici LGTM!

PR #1108, opened 2024-01-24
Expose CROSSPLANE_NAMESPACE to uptest

mbbush:
Since we started consuming https://github.com/upbound/uptest/pull/169, running `make uptest` or `make e2e` locally fails on the import step unless the CROSSPLANE_NAMESPACE env var is defined. This sets it in the uptest makefile step.

It would be great to backport this to at least 0.47, just to make local development on that branch easier, but I don't think I have permission to add the appropriate label.

I have:

Running uptest locally now passes without any additional env vars set.


ytsarev (approved):
Thanks for the catch, I hit the same in other Crossplane Configurations today and end up with the identical fix, so fully sure it's working.

ytsarev:
/test-examples="examples/docdb/cluster.yaml"

just in case, to have a centralized uptest run

PR #1111, opened 2024-01-26
Better detection of subpackages in apiVersion in uptest manifests

mbbush:
This allows resources in non-provider api groups to be specified if necessary.

I was trying to use a crossplane Usage resource in uptest, and it failed because the existing regex tried to figure out which provider-aws subpackage the crossplane api group was in, and failed. This change would remove one of the barriers to supporting more resource types in an uptest example, if that's necessary.
I have:

It hasn't yet, that's why this is a draft.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@
 PROVIDER_NAME := aws
 PROJECT_NAME := provider-$(PROVIDER_NAME)
 PROJECT_REPO := github.com/upbound/$(PROJECT_NAME)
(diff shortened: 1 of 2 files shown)


mbbush:
/test-examples="examples/iam/v1beta1/role-with-drc.yaml"

mbbush:
Uptest failed on an example where I added a DeploymentRuntimeConfiguration because it was waiting for the `Test` condition on the non-managed resource, so there's still some work to be done to allow the use of non-managed resources in uptest manifests.

PR #1116, opened 2024-01-30
[Breaking] Update external name config of two resources in cognitoidp

mbbush:

Both UserPoolDomain and UserPoolUiCustomization in cognitoidp have a terraform id that is completely knowable from their `spec.forProvider`, yet they still use the `config.IdentifierFromProvider` external name config. The result of this is that in order to assume control of an existing resource, you must specify the external-name annotation, and if you don't, you get an error when trying to create the resource. This doesn't seem in keeping with the XRM, where you declare the desired final state and let the provider worry about how to get there. 

This is technically a breaking change, but in very minor ways. 

Is this the kind of thing that will be easy to automate with conversion webhooks? 

Fixes #

I have:


config/externalname.go
```diff
@@ -110,9 +110,9 @@ var NoForkExternalNameConfigs = map[string]config.ExternalName{
 	// us-west-2_abc123
 	"aws_cognito_user_pool": config.IdentifierFromProvider,
 	// auth.example.org
-	"aws_cognito_user_pool_domain": config.IdentifierFromProvider,
+	"aws_cognito_user_pool_domain": config.TemplatedStringAsIdentifier("", "{{ .parameters.domain }}"),
```
mbbush, line 113: I first tried using `config.ParameterAsIdentifier("domain")`, but that removed `domain` from the `spec.forProvider` entirely, moving it to `metadata.name`. Especially given that the domain can be a FQDN, that seemed a step too far.

Issue #1071, opened 2024-01-08
Elasticache ReplicationGroup fails when engineVersion is provided


zhodowanec:
### What happened?
ReplicationGroup resource fails to sync when a value for the `engineVersion` parameter is defined. 

Error message: 
- `'observe failed: cannot compute the instance diff: failed to get *terraform.InstanceDiff: error parsing old engine_version: Malformed version: '`

### How can we reproduce it?

ReplicationGroup

```yaml
apiVersion: elasticache.aws.upbound.io/v1beta1
kind: ReplicationGroup
metadata:
  name: <cache-name>
spec:
  forProvider:
    atRestEncryptionEnabled: true
    automaticFailoverEnabled: true
    description: <description>
    engine: Redis
    engineVersion: "6.2.6"
    nodeType: cache.r5.large
    numNodeGroups: 1
    parameterGroupName: 
    port: 6379
    region: <region>
    replicasPerNodeGroup: 1
    subnetGroupNameRef:
      name: <cache-name>
    transitEncryptionEnabled: false
  providerConfigRef:
    name: <provider-config>
```

SubnetGroup

```yaml
apiVersion: elasticache.aws.upbound.io/v1beta1
kind: SubnetGroup
metadata:
  name: <cache-name>
spec:
  forProvider:
    region: <region>
    subnetIds:
      - <subnet-id>
  providerConfigRef:
    name: <provider-config>
```

### What environment did it happen in?

* Crossplane Version: `v1.14.4`
* Provider Version: `v0.47.1`
* Kubernetes Version: `v1.27.8`
* Kubernetes Distribution: `EKS`

turkenf:
Hi @zhodowanec, 

Thank you for raising this, have you tried using different versions, do you get the same error?

zhodowanec:
> Hi @zhodowanec,
> 
> Thank you for raising this, have you tried using different engine versions (e.g. engineVersion: "7.1"), do you get the same error?

@turkenf I have tried `7.1` as an engineVersion value and get the same error.

jeff-jsq:
I also get the same error when specifying `engineVersion: "7.1"` on `v0.47.1`. I can confirm the same `ReplicationGroup` definition works on `v0.46.1` (haven't tried others).

turkenf:
Thank you for your answers, I can reproduce the issue in provider version `0.47.1` and as @jeff-jsq said the issue does not occur in provider version `v0.46.1`

[...]

PR #1119, opened 2024-01-31
Add an example manifest for ReplicationGroup.elasticache with "engineVersion"

ulucinar:

This PR adds a new example manifest for the `ReplicationGroup.elasticache` resource with an explicit `engineVersion` as a regression test for #1071. We keep the original example manifest, which does not mention an explicit engine version, and thus requires less maintenance on our side (e.g., when the explicitly referenced engine version is no longer available).

I have:

Via uptest runs:
- Manifest with engine version: https://github.com/upbound/provider-aws/actions/runs/7730125050
- Generic manifest: https://github.com/upbound/provider-aws/actions/runs/7730127617

### DIFF
--- a/examples/elasticache/v1beta1/replicationgroup-engineVersion.yaml
+++ b/examples/elasticache/v1beta1/replicationgroup-engineVersion.yaml
@@ -9,12 +9,11 @@ metadata:
   name: example-${Rand.RFC1123Subdomain}
 spec:
   forProvider:
+    engineVersion: "7.1"
     automaticFailoverEnabled: true
     description: example description
     nodeType: cache.m4.large
     numCacheClusters: 2
+    parameterGroupName: default.redis7
     port: 6379
-    preferredCacheClusterAzs:
-      - us-west-1a
-      - us-west-1b
     region: us-west-1
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/aws/smithy-go v1.19.0
 	github.com/crossplane/crossplane-runtime v1.15.0-rc.0.0.20231215091746-d23a82b3a2f5
 	github.com/crossplane/crossplane-tools v0.0.0-20230925130601-628280f8bf79
-	github.com/crossplane/upjet v1.1.0-rc.0.0.20240130134554-62169a1988bf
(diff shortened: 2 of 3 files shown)


examples/elasticache/v1beta1/replicationgroup.yaml
```diff
@@ -13,7 +13,6 @@ spec:
     description: example description
     nodeType: cache.m4.large
     numCacheClusters: 2
-    parameterGroupName: default.redis7
```
ulucinar, line 16: Looks like the the parameter group's version should match the engine version. Removing this so that when version 8 is available, this parameter does not break the tests.

ulucinar:
/test-examples="examples/elasticache/v1beta1/replicationgroup-engineVersion.yaml"

examples/elasticache/v1beta1/replicationgroup-engineVersion.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: elasticache.aws.upbound.io/v1beta1
+kind: ReplicationGroup
+metadata:
+  annotations:
+    meta.upbound.io/example-id: elasticache/v1beta1/replicationgroup
+    uptest.upbound.io/timeout: "3600"
+  labels:
+    testing.upbound.io/example-name: example
+  name: example
```
turkenf, line 9: It would be better if we randomize the `metadata.name` fields.

```suggestion
  name: example-${Rand.RFC1123Subdomain}
```

turkenf (approved):
Thank you @ulucinar, LGTM.

examples/elasticache/v1beta1/replicationgroup-engineVersion.yaml
ulucinar: Good idea! 👍

ulucinar:
/test-examples="examples/elasticache/v1beta1/replicationgroup-engineVersion.yaml"

ulucinar:
/test-examples="examples/elasticache/v1beta1/replicationgroup.yaml"

ulucinar:
/test-examples="examples/iam/v1beta1/policy.yaml"

PR #1144, opened 2024-02-09
fix Secret/secretmanager.aws custom diff logic when replica config is empty

### LINKED ISSUE #1128: [Bug]: Secret never Sync

- [X] I have searched the existing issues

### Affected Resource(s)

secretsmanager.aws.upbound.io/v1beta1 - Secret

```
apiVersion: secretsmanager.aws.upbound.io/v1beta1
kind: Secret
metadata:
  name: test-dmaganto1
spec:
  forProvider:
    name: test-dmaganto1
    region: eu-central-1
    recoveryWindowInDays: 0
  providerConfigRef:
    name: aws-admin-my-account-upbound
```

### Steps to Reproduce

Apply this resource in the cluster

### What happened?

It never reach the state Sync: true

```shell
- lastTransitionTime: "2024-02-06T10:12:27Z"
    message: 'observe failed: cannot compute the instance diff: failed to compute
      the customized terraform.InstanceDiff: could not read replica block from config'
```

### Crossplane Version

1.14.5

### Provider Version

1.0.0

### Kubernetes Version

v1.28.2

### Kubernetes Distribution

EKS

### Additional Info

_No response_

erhancagirici:

Fixes the custom diff logic (introduced at https://github.com/upbound/provider-aws/pull/1107) for `replica` attribute when it is not specified in the config and the diff does not involve `replica` field. The custom diff function is triggered for any update diff, and the current logic tries to read the `replica` field unconditionally. When `replica` is not specified in config, the get operation fails the whole custom diff function, therefore the Observe operation.

This change fixes the custom diff logic by: 
- early exit with no error when there is no change/diff involved for `replica`. We only need to change
- early exit with no error if `replica` attibute is not specified in the MR spec. This is already handled properly by the regular diff and needs no customization.
 
Fixes #1128 

I have:

Tested with the spec at #1128

### DIFF
--- a/config/secretsmanager/config.go
+++ b/config/secretsmanager/config.go
@@ -35,13 +35,25 @@ func Configure(p *config.Provider) { //nolint:gocyclo
 				return diff, nil
 			}
 
+			resData, err := schema.InternalMap(r.TerraformResource.SchemaMap()).Data(state, diff)
+			if err != nil {
+				return nil, errors.New("could not construct resource data")
+			}
+
+			// do not customize diff if replica field has no change
+			if !resData.HasChange("replica") {
+				return diff, nil
+			}
 			currentReplicaSet, ok := r.TerraformResource.Data(state).Get("replica").(*schema.Set)
 			if !ok {
 				return nil, errors.New("could not read \"replica\" from state")
 			}
+
 			nisReplica, ok := config.Get("replica")
 			if !ok {
-				return nil, errors.New("could not read replica block from config")
+				// config is empty for replica, no need for custom diff logic
+				// this is already handled correctly with the built-in diff logic
+				return diff, nil
 			}
 			desiredReplicaList := nisReplica.([]interface{})
 			// this is the hash implementation of *schema.Set, which is unexported
--- a/examples/secretsmanager/v1beta1/secret-withreplica.yaml
+++ b/examples/secretsmanager/v1beta1/secret-withreplica.yaml
@@ -4,11 +4,13 @@ metadata:
   name: example-withreplica
   annotations:
     meta.upbound.io/example-id: secretsmanager/v1beta1/secret
+    uptest.upbound.io/update-parameter: '{"tags":{"updated-by":"crossplane"}}'
   labels:
     testing.upbound.io/example-name: secretsmanager
 spec:
   forProvider:
     name: example-withreplica-${Rand.RFC1123Subdomain}
+    recoveryWindowInDays: 0
     region: us-west-1
     replica:
       - region: us-west-2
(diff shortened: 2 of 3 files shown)


mbbush:
/test-examples="examples/secretsmanager/v1beta1/secret.yaml"

mbbush:
@erhancagirici this looks like a good fix, but in the interest of improving our test coverage, I'm puzzled why uptest doesn't fail on `main` without this fix. 

When I tried to reproduce the issue, I got the error described in #1128 when I was observing an existing secret, by setting the external name annotation, but not when I just created a new secret. It seems like that _should_ be caught by the import step in uptest, yet that passes consistently for me. 

Any ideas why?

mbbush:
It looks like the key difference in reproducing this bug is that when I duplicated the existing resource, I changed its metadata.name, which changed one of the crossplane-managed tags, which created a non-nil diff. 

This makes me want to get Update tests actually running in uptest for provider-aws.

config/secretsmanager/config.go
```diff
@@ -35,13 +35,25 @@ func Configure(p *config.Provider) { //nolint:gocyclo
 				return diff, nil
 			}
 
+			resData, err := schema.InternalMap(r.TerraformResource.SchemaMap()).Data(state, diff)
+			if err != nil {
+				return nil, errors.New("could not construct resource data")
```
mbbush, line 40: Is this really something we want to fail the whole observe loop for? I don't fully understand when this could happen, so I don't know how bad it is.

mbbush (commented):
I was able to reproduce the bug using uptest, and opened #1154 with examples that fail on main, but pass on this branch. 

I'd be happy for you to just cherry-pick/copy in that commit to your branch, or merge my PR and rebase, or whatever other git workflow you find easiest.

config/secretsmanager/config.go
erhancagirici: under "normal" operation this should never return an error. An error here indicates an issue with the resource schema or the produced state. More specifically, this is a very fundamental operation and the same line is also run by the terraform SDK library already in earlier stages of observe, so any error should be already emitted if any before visiting the custom diff function.So, IMHO it is OK to fail the whole observe here

erhancagirici:
> I was able to reproduce the bug using uptest, and opened #1154 with examples that fail on main, but pass on this branch.
> 
> I'd be happy for you to just cherry-pick/copy in that commit to your branch, or merge my PR and rebase, or whatever other git workflow you find easiest.

Many thanks 🎉  
As you have already discovered, this bug is reproducible when 
- a non-nil Update diff occurs (non-create, non-destroy)
- `replica` is not specified in resource config

I was also just about to send a very similar example change, and for the sake of completeness of this PR, I will cherrypick your commit here. Thanks again for the valuable contribution!

erhancagirici:
/test-examples="examples/secretsmanager/v1beta1/secret.yaml"

erhancagirici:
/test-examples="examples/secretsmanager/v1beta1/secret-withreplica.yaml"

sergenyalcin (approved):
Thanks @erhancagirici LGTM!

PR #1148, opened 2024-02-12
add web identity token configuration to ProviderConfig spec

### LINKED ISSUE #1007: WebIdentity credential source relies on eks.amazonaws.com/role-arn ServiceAccount annotation
### What happened?
Experimenting with having a single provider-aws-s3 assume different AWS IAM roles from different ProviderConfig objects, I noticed that using the WebIdentity credential source somewhat unintuitively depends on having the ServiceAccount referenced in the provider Deployment having the `eks.amazonaws.com/role-arn` annotation.

My understanding is that the reason for this is the following: when a Pod is being created that references a service account with the `eks.amazonaws.com/role-arn` annotation, some EKS internal mutating admission controller will modify the Pod object in the following way:

1.  add a projected volume with a serviceAccountToken block with an audience set to `sts.amazonaws.com`
2.  add a mount point in all containers referencing the volume in 1.
3.  set the `AWS_WEB_IDENTITY_TOKEN_FILE` environment variable to the token file in on the mount from 2.

If the annotation on the ServiceAccount is not found, the provider-aws-s3 will attempt to open the token file empty string (since looking up a nonexistent environment variable will not fail but return an empty string), normalized to `.`, which fails with the error message attached at the bottom of this report.

### How can we reproduce it?
By creating an provider-aws-s3 provider and a default ProviderConfig using a WebIdentity credential source, without any DeploymentRuntimeConfig containing a serviceAccountTemplate to add the `eks.amazonaws.com/role-arn` annotation and create an entity controlled by the provider (I tried with an S3 Object) referencing the mentioned ProviderConfig.

Inspecting the Pod object that the provider creates, it should be clear that the `AWS_WEB_IDENTITY_TOKEN_FILE` is not set, and attempting to act on the entity will cause the CannotConnectToProvider event detailed below to be emitted.

### Suggested fixes

The proper fix to my mind would be to modify the Deployment spec for all providers to include a service account token projection as this is a [publicly available feature](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/) of all supported versions of Kubernetes and modify the provider code to use this token.

As a temporary stop gap measure the docs could be updated to detail that the ServiceAccount annotation hack is currently needed.

* Crossplane Version: v1.14.3
* Provider Version: v0.45.0
* Kubernetes Version: v1.27.7-eks-4f4795d

The specific CannotConnectToProvider event associated with an s3 Object triggering this error:

```  Warning  CannotConnectToProvider  0s (x5 over 13s)  managed/s3.aws.upbound.io/v1beta1, kind=object  cannot initialize the no-fork async external client: cannot get terraform setup: cannot get account id: failed to retrieve aws credentials from aws config: failed to refresh cached credentials, failed to retrieve jwt from provide source, unable to read file at .: read .: is a directory```

### PR DESCRIPTION
**Pull request by erhancagirici**
Adds support for Web Identity token configuration from `ProviderConfig`
Fixes #1007

Currently Web Identity authentication configuration is only possible with `AWS_WEB_IDENTITY_TOKEN_FILE` and `AWS_ROLE_ARN` environment variables set. 
This causes confusion and ambiguity in several ways and has limitations:
- Although consumers specify the role to assume for Web Identity auth at `spec.credentials.webIdentity.roleARN`, they still need to specify the same role ARN at `AWS_ROLE_ARN` environment variable to make it work. As reported in #1007, this seems like the users depend on IRSA annotations, however it is actually a dependency to `AWS_ROLE_ARN` and `AWS_WEB_IDENTITY_TOKEN_FILE` environment variables. When consumers do IRSA config, these env vars are automatically set, which satisfies the config.
- The token to use for Web Identity cannot be specified via ProviderConfig spec and can only be specified through `AWS_WEB_IDENTITY_TOKEN_FILE`
  - Consumers might forgot to set the `AWS_ROLE_ARN` which emits uninformative errors.
  - This results in a heterogeneous config UX, where role ARN is provided through `ProviderConfig` and token through env.
  - Consumers need to prepare `ControllerConfig`s (deprecated) or `DeploymentRuntimeConfig`s to mount token files and inject env vars to provider pods.
  - Consumers are limited to a single Web Identity config per provider pod. It is not possible to configure tokens & roleARNs per provider config, where consumers want multiple provider configs referring to different Web Identity configs (as reported in https://github.com/upbound/provider-aws/issues/1007#issuecomment-1851229653)

This change expands `ProviderConfig` API specification with `spec.credentials.webIdentity.tokenConfig`, with options to reference a secret or filesystem location for the token to be used for `AssumeRoleWithWebIdentity` operations.

Consumers can specify the token's source (`Secret` or `Filesystem`) and set `secretRef` or `fs` respectively. 

> ℹ️   The change is backward compatible for consumers relying on old behavior where they set `AWS_WEB_IDENTITY_TOKEN_FILE` and `AWS_ROLE_ARN` environment variables, to allow a smooth transition
>
>⚠️ However, this approach is deprecated now and consumers are suggested to update their provider configs to specify the `tokenConfig`.

Example Web identity token configuration with Web Identity Token from a secret:

```yaml
kind: ProviderConfig
metadata:
  name: webidentity-example
spec:
  credentials:
    source: WebIdentity
    webIdentity:
      roleARN: arn:aws:iam::123456789012:role/providerexamplerole
      tokenConfig:
        source: Secret
        secretRef:
          key: token
          name: example-web-identity-token-secret
          namespace: upbound-system
```
example using a filesystem location. Note that `Filesystem` source option still needs the token file mounted to the filesystem of the provider pod.
The difference is that new API allows specifying the token file per `ProviderConfig`, meaning that you can mount multiple tokens to the provider pod and reference the desired one at each `ProviderConfig`
```yaml
kind: ProviderConfig
metadata:
  name: webidentity-example
spec:
  credentials:
    source: WebIdentity
    webIdentity:
      roleARN: arn:aws:iam::123456789012:role/providerexamplerole
      tokenConfig:
        source: Filesystem
        fs:
          path: /path/to/token/file
```

I have:

Tested with provided examples in the PR description for the new feature and uptest for regressions


erhancagirici:

erhancagirici:

erhancagirici:
Manually tested:
 - web identity with new secret ref token config
     - valid and invalid token paths
 - web identity with new filesystem ref token config
   - valid and invalid paths and tokens

 - old approach with environment variables `AWS_WEB_IDENTITY_TOKEN_FILE` and `AWS_ROLE_ARN`
 - non-web identity methods (for regressions)
    - access key and secret
    - IRSA
    - Upbound
- combinations with AssumeRoleChain added  

Upbound SaaS is also tested by @turkenf

internal/clients/aws.go
```diff
@@ -113,8 +115,22 @@ func pushDownTerraformSetupBuilder(ctx context.Context, c client.Client, pc *v1b
...
-			keyWebIdentityTokenFile: os.Getenv(envWebIdentityTokenFile),
+			keyRoleArn: aws.ToString(pc.Spec.Credentials.WebIdentity.RoleARN),
+		}
+		if pc.Spec.Credentials.WebIdentity.TokenConfig != nil {
+			tokenSelector := xpv1.CommonCredentialSelectors{
+				Fs:        pc.Spec.Credentials.WebIdentity.TokenConfig.Fs,
+				SecretRef: pc.Spec.Credentials.WebIdentity.TokenConfig.SecretRef,
+			}
+			tokenSource := pc.Spec.Credentials.WebIdentity.TokenConfig.Source
+			creds, err := resource.CommonCredentialExtractor(ctx, tokenSource, c, tokenSelector)
```
ulucinar, line 126: nit:
```suggestion
			creds, err := resource.CommonCredentialExtractor(ctx, pc.Spec.Credentials.WebIdentity.TokenConfig.Source, c, tokenSelector)
```

internal/clients/provider_config.go
```diff
@@ -318,6 +320,35 @@ func GetAssumeRoleWithWebIdentityConfig(ctx context.Context, cfg *aws.Config, we
 	return &awsConfig, nil
 }
 
+// GetAssumeRoleWithWebIdentityConfigViaTokenRetriever returns an aws.Config capable of doing
+// AssumeRoleWithWebIdentity using the token obtained from the supplied stscreds.IdentityTokenRetriever.
+func GetAssumeRoleWithWebIdentityConfigViaTokenRetriever(ctx context.Context, cfg *aws.Config, webID v1beta1.AssumeRoleWithWebIdentityOptions, tokenRetriever stscreds.IdentityTokenRetriever) (*aws.Config, error) {
+	stsclient := sts.NewFromConfig(*cfg, stsRegionOrDefault(cfg.Region)) //nolint:contextcheck
```
ulucinar, line 326: Why do we need to disable this linter?

internal/clients/provider_config.go
```diff
@@ -318,6 +320,35 @@ func GetAssumeRoleWithWebIdentityConfig(ctx context.Context, cfg *aws.Config, we
...
+				aws.ToString(webID.RoleARN),
+				tokenRetriever,
+				SetWebIdentityRoleOptions(webID),
+			)),
+		),
+	)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to load assumed role with web identity AWS config")
+	}
+	return &awsConfig, nil
```
ulucinar, line 343:
```suggestion
	return &awsConfig, errors.Wrap(err, "failed to assume role via web identity")
```

internal/clients/provider_config.go
```diff
@@ -318,6 +320,35 @@ func GetAssumeRoleWithWebIdentityConfig(ctx context.Context, cfg *aws.Config, we
...
+}
+
+// GetAssumeRoleWithWebIdentityConfigViaTokenFile returns an aws.Config capable of doing
+// AssumeRoleWithWebIdentity using a token file path.
+func GetAssumeRoleWithWebIdentityConfigViaTokenFile(ctx context.Context, cfg *aws.Config, webID v1beta1.AssumeRoleWithWebIdentityOptions, tokenFile string) (*aws.Config, error) {
+	return GetAssumeRoleWithWebIdentityConfigViaTokenRetriever(ctx, cfg, webID, stscreds.IdentityTokenFile(filepath.Clean(tokenFile)))
+}
```
ulucinar, line 350: nit: I think it makes sense to inline the inner function call:
```suggestion
```

internal/clients/provider_config.go
```diff
@@ -338,18 +369,72 @@ func UseDefault(ctx context.Context, region string) (*aws.Config, error) {
...
+	tokenSource   v1.CredentialsSource
+	tokenSelector v1.CommonCredentialSelectors
+}
+
+func (x *xpWebIdentityTokenRetriever) GetIdentityToken() ([]byte, error) {
+	token, err := resource.CommonCredentialExtractor(x.ctx, x.tokenSource, x.kube, x.tokenSelector)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not extract token from tokenSource")
+	}
+	return token, nil
```
ulucinar, line 384:
```suggestion
	return token, errors.Wrap(err, "could not extract token from tokenSource")
```

internal/clients/provider_config.go
```diff
@@ -338,18 +369,72 @@ func UseDefault(ctx context.Context, region string) (*aws.Config, error) {
...
+	if pcs.Credentials.WebIdentity.TokenConfig == nil {
+		cfg, err := UseDefault(ctx, region)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to get default AWS config")
+		}
+		return GetAssumeRoleWithWebIdentityConfig(ctx, cfg, *pcs.Credentials.WebIdentity, os.Getenv(envWebIdentityTokenFile))
+	}
+
+	// new behavior with tokenConfig in spec.credentials.webIdentity.tokenConfig
+
```
ulucinar, line 407:
```suggestion
```

internal/clients/provider_config.go
```diff
@@ -338,18 +369,72 @@ func UseDefault(ctx context.Context, region string) (*aws.Config, error) {
...
+	// the new behavior with tokenConfig does not rely on
+	// the AWS environment variables AWS_WEB_IDENTITY_TOKEN_FILE and AWS_ROLE_ARN.
+	// However, we start by constructing a default AWS config and
+	// AWS SDK enforces that when AWS_WEB_IDENTITY_TOKEN_FILE environment variable is set AWS_ROLE_ARN must be present.
+	// Otherwise, constructing the default AWS config fails.
+	// Hence, either both env vars must be set ( to support the case where the controller pod has extra AWS IRSA config
+	// which should be automatically injected  AWS_WEB_IDENTITY_TOKEN_FILE and AWS_ROLE_ARN environment variables already)
```
ulucinar, line 414: Could you please break these comments into at most 80 chars per line?

ulucinar (commented):
Thanks @erhancagirici for working on the ProviderConfig. Left some comments for you to consider.

internal/clients/provider_config.go
erhancagirici: Actually, I added this this function and did not use it after. My initial plan was to refactor GetAssumeRoleWithWebIdentityConfig, but then I decided not to do it. I'll remove this function.

erhancagirici:

erhancagirici:

ulucinar (approved):
Thanks @erhancagirici, lgtm.

PR #1162, opened 2024-02-16
Add support for msk connect resources

### LINKED ISSUE #374: Moving location(1), mskconnect(3) resources to v1beta1 version
Moving `location` (1) resource to v1beta1 version
- [ ] [`aws_location_map`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/location_map)

Moving `mskconnect` (3) resources to v1beta1 version
- [ ] [`aws_mskconnect_connector`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/mskconnect_connector)
- [ ] [`aws_mskconnect_custom_plugin`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/mskconnect_custom_plugin)
- [ ] [`aws_mskconnect_worker_configuration`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/mskconnect_worker_configuration)

### LINKED ISSUE #1146: Request for `aws_mskconnect_*` resources

**Terraform Resource Name:** 
aws_mskconnect_connector
aws_mskconnect_custom_plugin
aws_mskconnect_worker_configuration

### What is your use case?
Use of Debezium source/S3 sink connectors

mbbush:

Adds support for the three Msk Connect resources. 

Note: AWS has very limited api support for these resources. Other than the `scaling` object, every field on the terraform schema has `ForceNew: true`, so from the crossplane perspective, the resources are mostly immutable once created.  If you want to make a change, you'll have to delete it and make a new one. 

The WorkerConfiguration also has no delete method in the aws API, so once you've created a config with a given name, that's it, forever. 

There is some surprising lifecycle behavior around the Connector resource, that I'm not sure if it's possible to resolve. 

The AWS api exposes the `State` of a connector, which can be either "Creating", "Running", "Deleting" or "Failed". The terraform provider uses this state to determine whether a creation call succeeds, but does not expose it in the observed `status` as a property of the resource, so I can't think of a way to get it into crossplane. 

When I create a Connector resource that is broken (as in the `connector-nokafka.yaml` example, where the kafka cluster doesn't exist), what happens is that the create call runs for just over 20 minutes (I had to increase the timeout slightly from the 20 minutes configured in the terraform provider), and then fails, and sets the following conditions:
```
  - lastTransitionTime: "2024-02-17T23:47:54Z"
    reason: ReconcileSuccess
    status: "True"
    type: Synced
  - lastTransitionTime: "2024-02-17T23:47:54Z"
    reason: Creating
    status: "False"
    type: Ready
  - lastTransitionTime: "2024-02-18T00:08:23Z"
    message: 'async create failed: failed to create the resource: [{0 waiting for
      MSK Connect Connector (arn:aws:kafkaconnect:us-east-2:905418119848:connector/connector-broken/4f63f714-30c2-4dd7-bf0c-edced7778d4e-2)
      create: unexpected state ''FAILED'', wanted target ''RUNNING''. last error:
      UnknownError.Unknown: The last operation failed. Retry the operation.  []}]'
    reason: AsyncCreateFailure
    status: "False"
    type: LastAsyncOperation
```
The provider then immediately runs an Observe, which finds that the connector does exist, and that its `status` matches the `spec` (since the `State` is not exposed in the terraform schema), and decides that creation must have succeeded after all, filling in `status.atProvider` and setting the following conditions:
```
  - lastTransitionTime: "2024-02-17T23:47:54Z"
    reason: ReconcileSuccess
    status: "True"
    type: Synced
  - lastTransitionTime: "2024-02-18T00:08:25Z"
    reason: Available
    status: "True"
    type: Ready
  - lastTransitionTime: "2024-02-18T00:08:25Z"
    reason: Success
    status: "True"
    type: LastAsyncOperation
  - lastTransitionTime: "2024-02-18T00:08:25Z"
    reason: UpToDate
    status: "True"
    type: Test
```

I'm not quite sure how best to reconcile this behavior of the aws resource, and the terraform provider, with what crossplane expects. 

Fixes #374 
Fixes #1146 

I have:

Manually, in kind, using the manifests I committed. I was able to get the connector to enter a Running state, in which it did nothing because there was no traffic, and no actual working code to run, but the infrastructure was all healthy.

I'll run uptest on the manifests that don't require manual intervention.

Output from local run with the `connector.yaml` example:
```
NAME                                       READY   SYNCED   EXTERNAL-NAME   AGE
group.cloudwatchlogs.aws.upbound.io/test   True    True     op-vggf5w1o     154m

NAME                                           READY   SYNCED   EXTERNAL-NAME       AGE
securitygrouprule.ec2.aws.upbound.io/egress    True    True     sgrule-2572981504   154m
securitygrouprule.ec2.aws.upbound.io/ingress   True    True     sgrule-2191228865   154m

NAME                                  READY   SYNCED   EXTERNAL-NAME          AGE
securitygroup.ec2.aws.upbound.io/sg   True    True     sg-0b03fcec08eeca5d9   154m

NAME                                   READY   SYNCED   EXTERNAL-NAME              AGE
subnet.ec2.aws.upbound.io/subnet-az1   True    True     subnet-04e98dd98ea750f9f   154m
subnet.ec2.aws.upbound.io/subnet-az2   True    True     subnet-040975500b81b55f1   154m

NAME                         READY   SYNCED   EXTERNAL-NAME           AGE
vpc.ec2.aws.upbound.io/vpc   True    True     vpc-0319cb8653512d954   154m

NAME                                               READY   SYNCED   EXTERNAL-NAME                                                             AGE
deliverystream.firehose.aws.upbound.io/connector   True    True     arn:aws:firehose:us-east-2:226209437353:deliverystream/example-absdfsdf   154m

NAME                                READY   SYNCED   EXTERNAL-NAME   AGE
role.iam.aws.upbound.io/connector   True    True     connector       154m

NAME                                   READY   SYNCED   EXTERNAL-NAME                                                                                 AGE
cluster.kafka.aws.upbound.io/example   True    True     arn:aws:kafka:us-east-2:226209437353:cluster/example/d9f0e042-f520-4e34-a6d6-1ac5fce3e82d-4   154m

NAME                                         READY   SYNCED   EXTERNAL-NAME                                                                                       AGE
configuration.kafka.aws.upbound.io/example   True    True     arn:aws:kafka:us-east-2:226209437353:configuration/example/eae7df28-446c-4c93-9ecb-ab0737de0e01-4   154m

NAME                                                     READY   SYNCED   EXTERNAL-NAME                            AGE
connector.kafkaconnect.aws.upbound.io/connector          True    True     c6c84280-7796-42cc-9f4b-42607692741a-2   29m

NAME                                               READY   SYNCED   EXTERNAL-NAME                            AGE
customplugin.kafkaconnect.aws.upbound.io/example   True    True     29fec958-2694-4251-9a62-b2f7de64f7fc-2   154m

NAME                                                        READY   SYNCED   EXTERNAL-NAME                            AGE
workerconfiguration.kafkaconnect.aws.upbound.io/connector   True    True     2d5604c3-ec17-4394-be2f-8b07131963c3-2   154m

NAME                                 READY   SYNCED   EXTERNAL-NAME   AGE
bucket.s3.aws.upbound.io/connector   True    True     op-bonq5b6d     154m

NAME                               READY   SYNCED   EXTERNAL-NAME   AGE
object.s3.aws.upbound.io/example   True    True     empty.zip       154m
```


mbbush:
/test-examples="examples/kafkaconnect/v1beta1/customplugin.yaml"

mbbush:
/test-examples="examples/kafkaconnect/v1beta1/workerconfiguration.yaml"

mbbush:
/test-examples="examples/kafkaconnect/v1beta1/connector-broken.yaml"

mbbush:
/test-examples="examples/kafkaconnect/v1beta1/connector-broken.yaml"

The previous run actually failed because of the aws account hitting a VPC limit. It would be great if someone from upbound could submit some AWS quota increase requests for the number of VPCs per region.

mbbush:
/test-examples="examples/kafkaconnect/v1beta1/connector-broken.yaml"

mbbush:
The previous run failed with 
```
2024-03-01T19:10:45.0295557Z     logger.go:42: 19:10:44 | case/0-apply |       message: 'connect failed: cannot initialize the Terraform plugin SDK async external
2024-03-01T19:10:45.0296976Z     logger.go:42: 19:10:44 | case/0-apply |         client: cannot get terraform setup: cannot get referenced Provider: default:
2024-03-01T19:10:45.0298532Z     logger.go:42: 19:10:44 | case/0-apply |         Timeout: failed waiting for *v1beta1.ProviderConfig Informer to sync'
```
I haven't seen this particular failure before, but it doesn't seem at all related to this PR. 

Maybe it will pass a second time? 


/test-examples="examples/kafkaconnect/v1beta1/connector-broken.yaml"

mbbush:
The previous failure was caused by https://github.com/upbound/universal-crossplane/issues/442. Pinning the UXP version in the makefile resolved the issue.

/test-examples="examples/kafkaconnect/v1beta1/connector-nokafka.yaml"

mbbush:
Uptest is failing because of too many VPCs. This is a problem in the upbound AWS account, which I don't have access to. 

@jeanduplessis Could you please take a look at why cloud-nuke doesn't seem to be cleaning them up? Requesting a limit increase from AWS on the number of VPCs per region would also be very helpful.

jeanduplessis:
/test-examples="examples/kafkaconnect/v1beta1/connector-nokafka.yaml"

jeanduplessis:
@mbbush its should be fine now again. Ping me directly if the uptest run fails again with that issue.

mbbush:
/test-examples="examples/kafkaconnect/v1beta1/connector-nokafka.yaml"

Makefile
```diff
@@ -12,6 +12,8 @@ export TERRAFORM_PROVIDER_REPO ?= https://github.com/hashicorp/terraform-provide
 export TERRAFORM_DOCS_PATH ?= website/docs/r
 export PROVIDER_NAME
 
+export UXP_VERSION := 1.14.6-up.1
```
turkenf, line 15: Since the UXP version is currently [pinned](https://github.com/upbound/provider-aws/blob/9bc813763cb4ec29ca464cb375af530f259c5abc/Makefile#L71) in the main branch, you can continue by rebase after removing this line.

mbbush: Done.

mbbush:
I believe this is ready for review. @turkenf do you think you could take a look before the next provider-aws release?

I don't see any path forward to improving the behavior of the provider when creating a Connector results in creating a connector in state Failed in aws. The state is not exposed by the tf provider, so the only way I could even possibly query it would be to extract the aws client from the provider and make an explicit API call. 

I tried that, and was able to extract the aws client, but it was wrapped in a different type, and I couldn't figure out how to extract the underlying aws sdk client (it's probably possibly, I just don't know how). But regardless of whether it's possible I'm not sure if it's wise. 

My preference would be to merge this as-is, and release it, with the known issue that the provider does not expose error states with the connector. If there's consensus that it is wise to somehow add the connector's status field to the schema by invoking the SDK client directly, then I can work on that for a future release, as it would be a non-breaking change. 

Nothing significant has changed, so I'm confident the e2e tests will still pass. I'll rerun the ones that are testable automatically, but the full example which requires manual intervention is kind of a pain to run and takes a very long time, so I'd rather only do it one more time after this is otherwise approved.

mbbush:
/test-examples="examples/kafkaconnect/v1beta1/customplugin.yaml"

mbbush:
/test-examples="examples/kafkaconnect/v1beta1/connector-nokafka.yaml"

jeanduplessis:
/test-examples="examples/kafkaconnect/v1beta1/connector-nokafka.yaml"

config/kafkaconnect/config.go
```diff
@@ -0,0 +1,57 @@
...
+// Configure adds configurations for the kafkaconnect group.
+func Configure(p *config.Provider) {
+	p.AddResourceConfigurator("aws_mskconnect_connector", func(r *config.Resource) {
+		// This will always refer to a Cluster in the kafka api group, if it refers to any managed resource at all,
+		// but which property from the status of that cluster to use depends on the authentication mechanism chosen.
+		delete(r.References, "kafka_cluster.apache_kafka_cluster.bootstrap_servers")
+		r.References["kafka_cluster.apache_kafka_cluster.vpc.security_groups"] = config.Reference{
+			TerraformName:     "aws_security_group",
+			SelectorFieldName: "SecurityGroupSelector",
+			RefFieldName:      "SecurityGroupRefs",
```
turkenf, line 19: Is there a particular reason why you use singular resource names when defining references here?

As I understand, more than one security group can be passed, so we may consider removing these definitions.

config/kafkaconnect/config.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+		r.References["kafka_cluster.apache_kafka_cluster.vpc.subnets"] = config.Reference{
+			TerraformName:     "aws_subnet",
+			SelectorFieldName: "SubnetSelector",
+			RefFieldName:      "SubnetRefs",
```
turkenf, line 24: The same situation applies here as with the `security_groups` reference above.

config/kafkaconnect/config.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+		r.References["log_delivery.worker_log_delivery.firehose.delivery_stream"] = config.Reference{
+			TerraformName: "aws_kinesis_firehose_delivery_stream",
+			Extractor:     `github.com/crossplane/upjet/pkg/resource.ExtractParamPath("name",true)`,
+		}
+		r.References["worker_configuration.arn"] = config.Reference{
+			TerraformName: "aws_mskconnect_worker_configuration",
+			Extractor:     common.PathARNExtractor,
+		}
+		r.References["plugin.custom_plugin.arn"] = config.Reference{
```
turkenf, line 40: Since this reference is clearly in the terraform example, it should come automatically. Is there a reason to add it again?

config/kafkaconnect/config.go
```diff
@@ -0,0 +1,57 @@
...
+			TerraformName: "aws_mskconnect_custom_plugin",
+			Extractor:     common.PathARNExtractor,
+		}
+		// References only work to string fields.
+		delete(r.References, "plugin.custom_plugin.revision")
+		r.UseAsync = true
+		r.OperationTimeouts.Create = 30 * time.Minute
+		r.MetaResource.Description += ` Changes to any parameter besides "scaling" will be rejected. Instead you must create a new resource.`
```
turkenf, line 48: It would be nice if you could add a comment line explaining why you added these configurations.

examples/kafkaconnect/v1beta1/connector-nokafka.yaml
```diff
@@ -0,0 +1,273 @@
...
+    kafkaCluster:
+      - apacheKafkaCluster:
+          - bootstrapServers: "localhost:9092"
+            vpc:
+              - securityGroupSelector:
+                  matchLabels:
+                    testing.upbound.io/example-name: connector-nokafka
+                subnetSelector:
+                  matchLabels:
+                    testing.upbound.io/example-name: connector-nokafka
```
turkenf, line 40: We have added two Subnet resources below. I have no idea how it behaves since I haven't tested it, but shouldn't we pass it separately like in the [Cluster.kafka](https://github.com/crossplane-contrib/provider-upjet-aws/blob/0631d73c9ad50555f78989f3f5d098e67e346123/examples/kafka/v1beta1/cluster.yaml#L16) resource?

And as I mentioned above in the reference definition section, I think we should use plural resource names.

```suggestion
                subnetsRefs:
                  - name: subnet-az1
                  - name: subnet-az2
```

examples/kafkaconnect/v1beta1/connector-nokafka.yaml
```diff
@@ -0,0 +1,273 @@
...
+  forProvider:
+    region: us-east-2
+    forceDestroy: true
+
+---
+apiVersion: s3.aws.upbound.io/v1beta1
+kind: Object
+metadata:
+  annotations:
+    meta.upbound.io/example-id: kafkaconnect/v1beta1/customplugin
```
turkenf, line 139:
```suggestion
    meta.upbound.io/example-id: kafkaconnect/v1beta1/connector
```

examples/kafkaconnect/v1beta1/connector-nokafka.yaml
```diff
@@ -0,0 +1,273 @@
...
+kind: Group
+metadata:
+  annotations:
+    meta.upbound.io/example-id: kafkaconnect/v1beta1/connector
+    crossplane.io/external-name: ${Rand.RFC1123Subdomain}
+  labels:
+    testing.upbound.io/example-name: connector-nokafka
+  name: test
+spec:
+  deletionPolicy: Orphan
```
turkenf, line 233: Why do we set `deletionPolicy: Orphan`?

examples/kafkaconnect/v1beta1/connector-nokafka.yaml
```diff
@@ -0,0 +1,273 @@
...
+  forProvider:
+    region: us-east-2
+    retentionInDays: 1
+
+---
+apiVersion: ec2.aws.upbound.io/v1beta1
+kind: SecurityGroupRule
+metadata:
```
turkenf, line 241:
```suggestion
metadata:
  annotations:
    meta.upbound.io/example-id: kafkaconnect/v1beta1/connector
```

examples/kafkaconnect/v1beta1/connector-nokafka.yaml
```diff
@@ -0,0 +1,273 @@
...
+    fromPort: 0
+    toPort: 0
+    protocol: "all"
+    securityGroupIdRef:
+      name: sg
+    type: ingress
+---
+apiVersion: ec2.aws.upbound.io/v1beta1
+kind: SecurityGroupRule
+metadata:
```
turkenf, line 259:
```suggestion
metadata:
  annotations:
    meta.upbound.io/example-id: kafkaconnect/v1beta1/connector
```

examples/kafkaconnect/v1beta1/connector.yaml
```diff
@@ -0,0 +1,381 @@
...
+    kafkaCluster:
+      - apacheKafkaCluster:
+          - bootstrapServers: "REPLACE-ME plaintext kafka broker string:9092"
+            vpc:
+              - securityGroupSelector:
+                  matchLabels:
+                    testing.upbound.io/example-name: connector
+                subnetSelector:
+                  matchLabels:
+                    testing.upbound.io/example-name: connector
```
turkenf, line 35:
```suggestion
                subnetsRefs:
                  - name: subnet-az1
                  - name: subnet-az2
```

examples/kafkaconnect/v1beta1/connector.yaml
```diff
@@ -0,0 +1,381 @@
...
+  forProvider:
+    region: us-east-2
+    forceDestroy: true
+
+---
+apiVersion: s3.aws.upbound.io/v1beta1
+kind: Object
+metadata:
+  annotations:
+    meta.upbound.io/example-id: kafkaconnect/v1beta1/customplugin
```
turkenf, line 181:
```suggestion
    meta.upbound.io/example-id: kafkaconnect/v1beta1/connector
```

examples/kafkaconnect/v1beta1/connector.yaml
```diff
@@ -0,0 +1,381 @@
...
+spec:
+  deletionPolicy: Orphan
+  forProvider:
+    region: us-east-2
+    retentionInDays: 1
+---
+apiVersion: kafka.aws.upbound.io/v1beta1
+kind: Cluster
+metadata:
+  annotations:
```
turkenf, line 283:
```suggestion
  annotations:
    meta.upbound.io/example-id: kafkaconnect/v1beta1/connector
```

examples/kafkaconnect/v1beta1/connector.yaml
```diff
@@ -0,0 +1,381 @@
...
+              - enabledInBroker: true
+    region: us-east-2
+    tags:
+      foo: bar
+
+---
+
+apiVersion: kafka.aws.upbound.io/v1beta1
+kind: Configuration
+metadata:
```
turkenf, line 332:
```suggestion
metadata:
  annotations:
    meta.upbound.io/example-id: kafkaconnect/v1beta1/connector
```

examples/kafkaconnect/v1beta1/connector.yaml
```diff
@@ -0,0 +1,381 @@
...
+    name: example
+    region: us-east-2
+    serverProperties: |
+      auto.create.topics.enable = true
+      delete.topic.enable = true
+
+---
+apiVersion: ec2.aws.upbound.io/v1beta1
+kind: SecurityGroupRule
+metadata:
```
turkenf, line 349:
```suggestion
metadata:
  annotations:
    meta.upbound.io/example-id: kafkaconnect/v1beta1/connector
```

examples/kafkaconnect/v1beta1/connector.yaml
```diff
@@ -0,0 +1,381 @@
...
+    fromPort: 0
+    toPort: 0
+    protocol: "all"
+    securityGroupIdRef:
+      name: sg
+    type: ingress
+---
+apiVersion: ec2.aws.upbound.io/v1beta1
+kind: SecurityGroupRule
+metadata:
```
turkenf, line 367:
```suggestion
metadata:
  annotations:
    meta.upbound.io/example-id: kafkaconnect/v1beta1/connector
```

turkenf (commented):
Thank you for your effort in this PR @mbbush, I left a few initial review comments for you to consider.

config/kafkaconnect/config.go
mbbush: This is the standard I've observed in most cases with a plural reference field. Just parsing the names from an English perspective, a "Security group selector" doesn't give much of an indication one way or the other as to whether there's a single security group or multiple supported. 

Security Group Refs is grammatically correct, indicating multiple references to one security group each, while the default "Security Groups Ref" would imply one reference to multiple security groups. 

It would be great to have consistency on these sorts of names, and I thought this was following the existing prevailing standard, but I haven't gone through and looked at them more broadly to see how many multi-references use which patterns.

examples/kafkaconnect/v1beta1/connector-nokafka.yaml
mbbush: Because the target field of the reference is an array type, the selector finds multiple matches and happily resolves all of them. 

Here's what the spec looks like once references have been resolved:
```
         vpc:
          - securityGroupRefs:
            - name: sg
            securityGroupSelector:
              matchLabels:
                testing.upbound.io/example-name: connector-nokafka
            securityGroups:
            - sg-040d3c5051ff7055e
            subnetRefs:
            - name: subnet-az1
            - name: subnet-az2
            subnetSelector:
              matchLabels:
                testing.upbound.io/example-name: connector-nokafka
            subnets:
            - subnet-0c401af028b9bc05e
            - subnet-0002763a0a9e8db5b
```

config/kafkaconnect/config.go
mbbush: Looking through them more broadly wasn't too hard:
`git grep -A 4 'r.ResolveMultiple'`

In provider-upjet-aws, there are 232 multiple references, including the ones added in this PR. Only 36 contain the `sRefs` string in their target field name, from which I infer that the vast majority have been customized in this way, or have a non-plural terraform field name. 

I'd like to be able to make upjet smart enough to handle simple plurals, but in order for that to be a non-breaking change, we'd have to go manually configure those 36 examples to use the old style, and then all the other upjet-based providers would have to do the same in order to consume the new version. So it's probably not worth it. 

It would be great to be able to have a check for new field references, but maybe we can just enforce it manually in code review.

config/kafkaconnect/config.go
mbbush: Not a strong one, just my preference for being explicit over implicit. I wasn't looking at the examples when I was writing the config, I was just specifying all the references based on my understanding of the schema. 

My preference would be to leave this in, but if you'd rather I remove it, I will.

examples/kafkaconnect/v1beta1/connector-nokafka.yaml
mbbush: Leftover from manual testing where I wanted to see the actual logs to get the connector working. Will remove (from both examples)

config/kafkaconnect/config.go
turkenf: Thank you for the detailed explanation, then it seems logical to continue in its current form.

config/kafkaconnect/config.go
turkenf: https://github.com/crossplane-contrib/provider-upjet-aws/pull/1162#discussion_r1535511160

config/kafkaconnect/config.go
turkenf: We generally prefer not to define automatically defined references. I don't have a strong reason for it to be removed here either, so I leave the choice up to you.

turkenf (approved):
Thank you @mbbush, we can merge after resolving the conflicts.

mbbush:
@turkenf Rebased and resolved conflicts. It's not clear to me if you've already started the process of releasing v1.3.0. It would be great if this made it in, but it's not a huge problem for me if it doesn't.

turkenf:
@mbbush, we will include this PR.

mbbush:
Thanks! I'm rebasing/committing from the train on my way in to the office, so I didn't have the bandwidth to run `make generate` locally. But I applied the `diff` from the failed check-diff ci job using `patch`, so it should pass now.

turkenf:

PR #1194, opened 2024-03-06
Decrease linter's memory usage

ulucinar:

Depends on: https://github.com/upbound/uptest/pull/187

#### Historical Context & Problem Statement

The linter jobs in the `upbound/provider-aws` repository have been facing recurring failures. Initially, these failures were mitigated by switching to larger self-hosted runners (runners with the `e2-standard-8` labels in the `upbound` organization), but [the issues resurfaced](https://github.com/upbound/provider-aws/issues/737) due to a performance regression in the `musttag` linter. Subsequently, we upgraded the `golangci-lint` version to `v1.54.0` which resolved the performance regression with that specific linter. But later, we encountered further challenges, prompting a switch to an even larger runners. The runners we currently use only for the linter job with the label `Ubuntu-Jumbo-Runner`. Despite these adjustments, the linter jobs has started failing again, primarily due to the high peak memory consumption during linting, with cold analysis cache runs consuming over 50 GB of peak memory, depending on the number of available CPU cores.

#### Investigation & Findings
The substantial memory usage was traced back to the linter runner's analysis phase. We considered (and investigated some of) the following potential remediations for the linter issues we've been experiencing:

- Just configuring the `skip-files` to exclude, for instance, the generated files as we do right now for the official provider repos, does not help because it just skips reporting the issues. All the generated files are still analyzed and the processing needed to analyze the source code is the memory consuming part. So it does not help us.
- Tried bumping the linter runner to the latest version with a hope to enjoy any upstream improvements. The memory consumption is still high.
- Tried different sets of enabled linters & examined the resulting memory profiles. This has a potential but the processing needed for the analysis is common to the linters and probably defines a lower bound on the memory consumed. I've seen ~20 GB memory consumption even with a single linter configured. Configuring the linter runner only with the default set of linters (`errcheck`, `gosimple`, `govet` , `ineffassign`, `staticcheck`, `unused`) still consumes over 30 GB of memory. An idea could be to partition the current set of enabled linters across multiple jobs and hope no job will hit the memory limits.
- Employ the build tags that will make the analysis actually less costly in contrast to the` skip-files` argument. 
- Switch to even larger runners: Ideally we would like to avoid this. We will also be moving their repositories out of the `upbound` Github organization.
- Decrease the concurrency of the linter runner: The linter runner by default uses all the available logical cores available to it. Concurrency-limiting like we did for building & pushing the provider families to limit the load on our package registry may help here, in expense of increased runtime. The common processing phase in the runner prior to the analysis phase could be a limiting factor (please see above).
- Increase the pressure on the garbage collector: A common technique with the expense of increased CPU consumption and runtime.
- Disable certain linters: We are also trying to avoid this. But if we figure out a single linter that's breaking the memory limit, might be worth implementing.
- Break the family provider repos that need it: This is problematic because the resource providers of the same family are coupled (common provider configuration, cross-resource references ,etc.), although we had to decrease this coupling (specifically cross-resource reference induced coupling) for the multi-version CRD support. Also there's the maintenance overhead of the extra repositories. Furthermore, we can achieve a similar effect via the build tags (please see above).

#### Implemented Strategy in this PR
- Run the linter in two phases:
  1. Employ build constraints (tags) for the initial construction of a linter runner analysis cache and limit the concurrency to 1 when constructing this cache. We avoid the two hotspots in the `provider-aws` repo during cache construction: API registration in `apis/zz_register.go` & API group configuration in `config/provider.go`.
  2. After the cache construction, run the linters on the whole provider codebase with the maximum available concurrency.

The analysis cache expires in 7 days or when the module’s `go.sum` changes. If the analysis cache can successfully be restored, then the initial cache construction phase is skipped and just full linting with the maximum available concurrency is performed.

For generating the build constraints (tags), we use the [buildtagger](https://github.com/upbound/uptest/pull/187) tool. We currently don’t utilize the build tags for building the resource providers in isolation because:
- We currently don’t need this, i.e., the actual build process is not as resource hungry as linting
- It impacts the developer experience: It has implications in IDE settings, it will require changes to our build pipelines, etc.
- We would need to implement dependency resolution for cross-resource references and we’ve so far avoided this for the family providers (increased complexity), or, we would need to require the manual maintenance of dependencies (which is error prone).

#### Observed Improvements
In an example run with the two-phase strategy, the cache construction phase consumed a peak memory of ~13 GB and the full linting phase consumed a peak of 24,3 GB, which corresponds to a ~57% reduction in peak memory consumption compared to a single phase run of the linters on the same machine (an M2 Mac with 12 cores). The total execution time of both phases is ~14m, which is about the same time it takes the linters to run in a single phase (when we run the linters in a single phase on cold analysis cache, the peak memory consumption was ~57 GB and the execution time was ~14 min):

Here are results from example runs on an M2 Mac with 12 logical cores and 32 GB of physical memory:

1. Linter run in single phase (without the proposed initial cache construction phase) on cold analysis cache on the main branch’s Head (commit: `fb0fb486e6225cdab27a447c48cb36f98464884e`). Linter runner version: `v1.55.2`:
Average memory consumption is 40560.9MB, max is 58586.2MB. Execution took 13m49.188064208s.

1. Linter run in single phase with the analysis cache with the same parameters as above:
Average memory consumption is 104.5MB, max is 191.1MB. Execution took 7.325796125s.

1. Two-phase linter run example on cold analysis cache on the main branch’s Head (commit: `fb0fb486e6225cdab27a447c48cb36f98464884e`). Linter runner version is `v1.55.2`:
Average memory consumption in the cache construction phase is 9272.4MB and the peak consumption is 13301.6MB.
Execution of the first phase took 11m13.023597833s. For the second phase (full linting with all the available CPU cores), the average memory consumption is 9331.7MB and the peak is 24904.9MB. The execution of the second phase took 3m10.447865375s. 

The linter job now fits into a (standard) Github hosted runner with the label `ubuntu-22.04` ([with 16 GB of physical memory & 4 cores](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)). So, in preparation of moving `provider-aws` out of the `upbound` Github organization, this PR also changes the `lint` job's runner from `Ubuntu-Jumbo-Runner` to `ubuntu-22.04`.

#### Developer Experience
- API group configuration is implemented with slight differences:
  - The actual configuration still resides under `config/<API group prefix, e.g., acm>/config.go`. There are no upjet resource configuration API changes, i.e., the contents of `config.go` stays the same, and because the build tags for these files are automatically generated, there's no need to manually tag these files.
  - Registrations of the API group “Configure” functions are still done in `config/provider.go` but in a slightly different way. The new registration method is like follows:
```go
func init() {
   ProviderConfiguration.AddConfig(acm.Configure)
   ProviderConfiguration.AddConfig(acmpca.Configure)
   ProviderConfiguration.AddConfig(apigateway.Configure)
   …
```
  - `make lint` can be run with its old semantics. When the linter is running, the generated build tags will be observed in the source tree. If uninterrupted, these tags will be removed by the make target.
  - `make build` can be run with its old semantics & behavior

I have:

- The linter runner can successfully report issues: https://github.com/upbound/provider-aws/actions/runs/8196350610/job/22416386815?pr=1194
image

- If the linter runner's analysis cache is **not** restored, it will run to populate the cache without reporting any issues (even if there actually are issues): https://github.com/upbound/provider-aws/actions/runs/8196350610/job/22416386815?pr=1194
image

- If the linter runner's analysis cache could be successfully restored, then the cache population step is skipped: https://github.com/upbound/provider-aws/actions/runs/8196537292/job/22416975348?pr=1194
image

- Uptest run for `Cluster.eks`: https://github.com/upbound/provider-aws/actions/runs/8209193860


ulucinar:
/test-examples="examples/eks/v1beta1/cluster.yaml"

ulucinar:

Makefile
```diff
@@ -325,7 +341,24 @@ kustomize-crds: output.init $(KUSTOMIZE) $(YQ)
...
-.PHONY: kustomize-crds
+ifeq ($(RUN_BUILDTAGGER),true)
+lint.init: build-lint-cache
+lint.done: delete-build-tags
+
+build-lint-cache: $(GOLANGCILINT)
+	@$(INFO) Running golangci-lint with the analysis cache building phase.
+	@(BUILDTAGGER_DOWNLOAD_URL=$(BUILDTAGGER_DOWNLOAD_URL) ./scripts/tag.sh && \
+	(([[ "${SKIP_LINTER_ANALYSIS}" == "true" ]] && $(OK) "Skipping analysis cache build phase because it's already been populated") && \
+	[[ "${SKIP_LINTER_ANALYSIS}" == "true" ]] || $(GOLANGCILINT) run -v --build-tags account,configregistry,configprovider,linter_run -v --concurrency 1 --disable-all --exclude '.*')) || $(FAIL)
```
sergenyalcin, line 357: nit: Can we consider increasing the Concurrency?

Makefile
```diff
@@ -325,7 +341,24 @@ kustomize-crds: output.init $(KUSTOMIZE) $(YQ)
...
+	@$(OK) Running golangci-lint with the analysis cache building phase.
+
+delete-build-tags:
```
sergenyalcin, line 360: As far as I understand, this last operation allows us to remove the tags we added after the lint phase. In this context, the relevant tags are only modified while the linter runs. This is important so that the local env is not affected.

apis/linter_run.go
sergenyalcin, line 1: If I remember correctly, this file was brought up in the context of the fragmentation of the registration phase. It will be needed when we want to add all the resources to the schema. I can't find any use of this function in the repo. Am I missing something, or do we really need this file?

config/registry.go
sergenyalcin, line 1: If I understand correctly, we separate these two files to reduce the load of the linter. We mark the file where the configurations are registered (`config/provider.go`) with a tag that ignores linter. This way, while lintering some provider configuration functions in the `config/registry.go` (e.g., `getProviderSchema`, `GetProvider`), we leave only the configuration registration in `config/provider.go` and avoid the package loading overhead because of the many import statements there.

Actually, I'm writing this comment to understand if there is another area of improvement and purpose in this refactoring. If there is, we can consider doing it for other providers for compatibility.

sergenyalcin (commented):
Thanks @ulucinar for your great effort. Left a few comments.

Makefile
```diff
@@ -325,7 +340,18 @@ kustomize-crds: output.init $(KUSTOMIZE) $(YQ)
...
 
-.PHONY: kustomize-crds
+ifeq ($(RUN_BUILDTAGGER),true)
+lint.init: build-lint-cache
+
+build-lint-cache: $(GOLANGCILINT)
+	@$(INFO) Running golangci-lint with the analysis cache building phase.
+	@(BUILDTAGGER_DOWNLOAD_URL=$(BUILDTAGGER_DOWNLOAD_URL) ./scripts/tag.sh && \
+	(([[ "${SKIP_LINTER_ANALYSIS}" == "true" ]] && $(OK) "Skipping analysis cache build phase because it's already been populated") && \
+	[[ "${SKIP_LINTER_ANALYSIS}" == "true" ]] || $(GOLANGCILINT) run -v --build-tags account,configregistry,configprovider,linter_run -v --concurrency 1 --disable-all --exclude '.*')) || $(FAIL)
```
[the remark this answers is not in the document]
ulucinar: Done.

Makefile
```diff
@@ -48,6 +48,15 @@ GO_REQUIRED_VERSION ?= 1.21
 # Uncomment below if you need to override the version.
 GOLANGCILINT_VERSION ?= 1.55.2
 
+RUN_BUILDTAGGER ?= true
+# if RUN_BUILDTAGGER is set to "true", we will use build constraints
+# and use the buildtagger tool to generate the build tags.
+ifeq ($(RUN_BUILDTAGGER),true)
+GO_LINT_ARGS ?= -v --build-tags all
+BUILDTAGGER_VERSION ?= v0.12.0-rc.0.28.gdc5d6f3
+BUILDTAGGER_DOWNLOAD_URL ?= https://s3.us-west-2.amazonaws.com/upbound.official-providers-ci.releases/main/$(BUILDTAGGER_VERSION)/bin/$(SAFEHOST_PLATFORM)/buildtagger
```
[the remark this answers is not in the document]
ulucinar: We've switched recently from manual publishing via Github releases to automated publishing to the S3 buckets in the `upbound/uptest` repository with this PR: https://github.com/upbound/uptest/pull/183

Publishing via the S3 buckets is also common in the ecosystem, e.g., the `crossplane/crossplane` repo or the `upbound/up` repo also publish their CLI binaries to S3 buckets. The `upbound/build` Git submodule has specific targets to facilitate this.

We used to build & attach the `uptest` binaries (for different platforms) to the Github releases of the `upbound/uptest` repository but we've now split `uptest` & `crddiff` and we've added the `buildtagger` as well, which made releasing via Github even more difficult. We may consider actions to automate releases via Github but in line with the other repos in our ecosystem (and thus the build submodule), I would prefer sticking now with releasing via the cloud storage.

Makefile
ulucinar: Yes, correct. We run the `buildtagger` once more to remove the build tags. Another option would be to copy the code base to a staging folder and generate the build tags on that temporary location so that we don't need to remove the tags. But I preferred this approach because we would need to be careful about what gets copied into the temporary location. I've also updated the developer experience section that warns about the in-place modification during the `make lint` run. If this disturbs other developers, then we can consider linting on a repository copy.

apis/linter_run.go
ulucinar: As explained [here](https://golangci-lint.run/usage/faq/#why-do-you-have-typecheck-errors), the code that the linter runner is processing (analyzing) must compile. The API registration code (`apis/zz_register.go`) is one of the hot spots, which imports all the API packages and is thus costly to analyze. The `apis/linter_run.go` keeps the linter happy by providing a definition of `apis.AddToScheme` & `apis.AddToSchemes` and the original definitions from `apis/zz_register.go` are excluded by the supplied build constraints to the linter. So, the definitions in this file are only used by the linter and the actual build uses the definitions from `apis/zz_register.go`. 

This also means that the generated file `apis/zz_register.go` is not analyzed. We can in theory distribute API registration across the individual resource provider (API group) modules but the cross-resource references necessitate a form of dependency resolution as discussed under the `Implemented Strategy in this PR` section of the PR description. So avoiding these hot spots at the expense of not linting them is a trade-off we make here.

config/registry.go
ulucinar: Yes, this is the other hot spot which imports packages across the provider, and to decrease the memory consumption during the analysis cache construction, we also avoid these imports. Please also see the comment [here](https://github.com/upbound/provider-aws/pull/1194#discussion_r1519899246). Because this is not a generated file, I've preferred to split it contrary to what we've done with `apis/zz_register.go`. But the technique is the same as what we have done with the other hot spot.

apis/linter_run.go
sergenyalcin: Gotcha! Thanks for the clarification. This part can also be a bit tricky for other contributors. So, I am leaving this comment unresolved for documenting purposes.

Makefile
ulucinar: A concurrency of 4 for the initial analysis cache build phase has failed here:
https://github.com/upbound/provider-aws/actions/runs/8246986803/job/22554209938?pr=1194

ulucinar: [Here](https://github.com/upbound/provider-aws/actions/runs/8235027938/job/22518207849)'s a run with a concurrency of 1. It took ~21 min and ~14 GB of peak memory to complete the initial phase:
image

And [here](https://github.com/upbound/provider-aws/actions/runs/8247646010/job/22556266352?pr=1194)'s another run with a concurrency of 3. It took ~14 min ~19 GB of peak memory to complete the initial phase:
image

Let's switch to a concurrency of 3 for the initial phase (concurrency of 4 has failed). The second phase already uses the max available concurrency. 

If we observe any memory issues for the initial phase, we can decrease it later again.

PR #1204, opened 2024-03-12
Unify TF AWS provider configuration logic

erhancagirici:

refactors AWS client configuration logic with a single path.

Previously, the AWS config was following various paths due to:
- Terraform Provider Configuration does not support AssumeRoleChain with lengths > 1
- Terraform CLI was using the shared scheduler, which required configuration push-down
- `eks.clusterAuth` resource has a manual controller implementation, since it is not terraform-based
To handle these paths, a native AWS configuration was being constructed along with the terraform configuration attributes.
These were causing in a non-unified configuration path, and duplicated AWS STS API calls due to configuration being processed by the native method and Terraform provider configure for no-fork.

Since the TF CLI is removed all relevant configuration paths can be removed to simplify.

This PR:
- removal of Terraform configuration block preparation for CLI
  - TerraformSetup.Configuration is no more present
  - no push-down configuration
- AWS configuration and credential resolution is solely handled by native AWS config.
  - All auth methods resolves down to the final AccessKey and Secret key (temporary) credentials
  - other configuration such as AssumeRoleChain and custom endpoints are also handled by the AWS config
- move away from native TF Provider's Configure() method
  - this method had a potential to introduce unintended side effects and bugs, due to its singleton nature.
  - instead, the native TF configuration object is directly filled using the already-obtained native AWS config & credentials.
  - This reduces the extra AWS STS calls for GetCallerIdentity and AssumeRole operations.

As a result this leads to 50% reduction in STS calls made per reconcile. Further improvement to STS calls will follow with a [cache implementation](https://github.com/crossplane-contrib/provider-upjet-aws/issues/997#issuecomment-1991658640). 

I have:

Tested manually with following ProviderConfig specifications:
- Static creds from secret
- Static creds from secret with AssumeRoleChain
- WebIdentity
- WebIdentity with AssumeRoleChain 
- IRSA
- IRSA with AssumeRoleChain
- Upbound Auth  (thanks @turkenf for the test)

### DIFF
--- a/config/registry.go
+++ b/config/registry.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	// Note(ezgidemirel): we are importing this to embed provider schema document
 	_ "embed"
+
 	"github.com/crossplane/upjet/pkg/config"
 	"github.com/crossplane/upjet/pkg/registry/reference"
 	conversiontfjson "github.com/crossplane/upjet/pkg/types/conversion/tfjson"
--- a/config/registry.go
+++ b/config/registry.go
@@ -8,9 +8,6 @@ import (
 	"context"
 	// Note(ezgidemirel): we are importing this to embed provider schema document
 	_ "embed"
-	"reflect"
-	"unsafe"
-
 	"github.com/crossplane/upjet/pkg/config"
 	"github.com/crossplane/upjet/pkg/registry/reference"
 	conversiontfjson "github.com/crossplane/upjet/pkg/types/conversion/tfjson"
@@ -80,7 +77,7 @@ func getProviderSchema(s string) (*schema.Provider, error) {
 // configuration is being read for the code generation pipelines.
 // In that case, we will only use the JSON schema for generating
 // the CRDs.
-func GetProvider(ctx context.Context, generationProvider bool) (*config.Provider, *xpprovider.AWSClient, error) {
+func GetProvider(ctx context.Context, generationProvider bool) (*config.Provider, error) {
 	var p *schema.Provider
 	var fwProvider fwprovider.Provider
 	var err error
@@ -91,16 +88,7 @@ func GetProvider(ctx context.Context, generationProvider bool) (*config.Provider
 		fwProvider, p, err = xpprovider.GetProvider(ctx)
 	}
 	if err != nil {
-		return nil, nil, errors.Wrapf(err, "cannot get the Terraform provider schema with generation mode set to %t", generationProvider)
-	}
-	// we set schema.Provider's meta to nil because p.Configure modifies
-	// a singleton pointer. This further assumes that the
-	// schema.Provider.Configure calls do not modify the global state
-	// represented by the config.Provider.TerraformProvider.
-	var awsClient *xpprovider.AWSClient
-	if !generationProvider {
-		// #nosec G103
-		awsClient = (*xpprovider.AWSClient)(unsafe.Pointer(reflect.ValueOf(p.Meta()).Pointer()))
+		return nil, errors.Wrapf(err, "cannot get the Terraform provider schema with generation mode set to %t", generationProvider)
 	}
 
 	modulePath := "github.com/upbound/provider-aws"
@@ -131,7 +119,6 @@ func GetProvider(ctx context.Context, generationProvider bool) (*config.Provider
 			injectFieldRenamingConversionFunctions(),
 		),
 	)
-	p.SetMeta(nil)
 	pc.BasePackages.ControllerMap["internal/controller/eks/clusterauth"] = "eks"
 
 	for _, configure := range ProviderConfiguration {
@@ -139,7 +126,7 @@ func GetProvider(ctx context.Context, generationProvider bool) (*config.Provider
 	}
 
 	pc.ConfigureResources()
-	return pc, awsClient, nil
+	return pc, nil
 }
 
 // CLIReconciledResourceList returns the list of resources that have external
--- a/internal/clients/aws.go
+++ b/internal/clients/aws.go
@@ -28,7 +28,6 @@ const (
 
 type SetupConfig struct {
 	TerraformProvider *schema.Provider
-	AWSClient         *xpprovider.AWSClient
 }
 
 func SelectTerraformSetup(config *SetupConfig) terraform.SetupFn { // nolint:gocyclo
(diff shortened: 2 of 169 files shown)


config/registry.go
ulucinar, line 1: In case you've not seen it yet (as the linter is not working properly for a while), there are linter issues [here](https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/8263053130/job/22605446187?pr=1204):
image

```go
import (
	"context"
	// Note(ezgidemirel): we are importing this to embed provider schema document
	_ "embed"

	"github.com/crossplane/upjet/pkg/config"
	"github.com/crossplane/upjet/pkg/registry/reference"
	conversiontfjson "github.com/crossplane/upjet/pkg/types/conversion/tfjson"
	tfjson "github.com/hashicorp/terraform-json"
	fwprovider "github.com/hashicorp/terraform-plugin-framework/provider"
	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
	"github.com/hashicorp/terraform-provider-aws/xpprovider"
	"github.com/pkg/errors"

	"github.com/upbound/provider-aws/hack"
)
```

ulucinar:
The issue with the `publish-artifacts` is a known one with the recent SSO changes. cc. @turkenf

internal/clients/aws.go
```diff
@@ -230,18 +99,57 @@ func (m *metaOnlyPrimary) Meta() any {
...
-	p := *config.TerraformProvider
-	diag := p.Configure(context.WithoutCancel(ctx), &tfsdk.ResourceConfig{
-		Config: ps.Configuration,
+// configureNoForkAWSClient populates the supplied *terraform.Setup with
+// Terraform Plugin SDK style AWS client (Meta) and Terraform Plugin Framework
+// style FrameworkProvider
+func configureNoForkAWSClient(ctx context.Context, ps *terraform.Setup, config *SetupConfig, awsCfg *aws.Config, creds aws.Credentials, pc *v1beta1.ProviderConfig) error { //nolint:gocyclo
+	tfAwsConnsCfg := xpprovider.AWSConfig{
+		AccessKey:                     creds.AccessKeyID,
+		EC2MetadataServiceEnableState: imds.ClientDefaultEnableState,
```
ulucinar, line 108: Do we know the implications of this? A comment would help here. Have we preserved the old implementation?

internal/clients/aws.go
```diff
@@ -230,18 +99,57 @@ func (m *metaOnlyPrimary) Meta() any {
...
+				}
+			}
+		}
+	}
+
+	// only used for retrieving the ServicePackages from the singleton provider instance
+	p := config.TerraformProvider.Meta()
+	tfAwsConnsClient, diags := tfAwsConnsCfg.GetClient(ctx, &xpprovider.AWSClient{
+		// #nosec G103
+		ServicePackages: (*xpprovider.AWSClient)(unsafe.Pointer(reflect.ValueOf(p).Pointer())).ServicePackages,
```
ulucinar, line 139: If I remember correctly, this service packages concept played a role in properly handling the resource tags. Have you also tested if resource tags are still being handled properly, e.g., are the default tags and tag updates still working as expected?

ulucinar (commented):
Thank you very much @erhancagirici for cleaning up the AWS ProviderConfig code base and improving the situation with the number of per-MR STS calls the provider makes . Could you please also update the PR description with:
- The details of the changes this PR proposes
- The manual tests you did to verify these changes
- STS calls saved with these changes

We can also discuss briefly in the PR description the next iteration of planned improvements, where we would like to introduce a credential cache to further decrease the number of STS calls we make.

internal/clients/provider_config.go
```diff
@@ -139,6 +127,26 @@ func GetAWSConfig(ctx context.Context, c client.Client, mg resource.Managed) (*a
 	return SetResolver(pc, cfg), nil
 }
 
+// GetAWSConfigWithProviderUsage obtains the provider config referenced by the
+// specified managed resource and produces a config that can be used to
+// authenticate to AWS and tracks the ProviderConfigUsage. Useful for obtaining
+// AWS config for non-upjet based MR controllers.
+func GetAWSConfigWithProviderUsage(ctx context.Context, c client.Client, mg resource.Managed) (*aws.Config, error) { // nolint:gocyclo
```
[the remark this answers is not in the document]
erhancagirici: This seems to be a pre-refactoring leftover. good catch, thanks!

internal/clients/aws.go
```diff
@@ -25,39 +23,26 @@ import (
...
+		t := resource.NewProviderConfigUsageTracker(c, &v1beta1.ProviderConfigUsage{})
+		if err := t.Track(ctx, mg); err != nil {
+			return terraform.Setup{}, errors.Wrap(err, "cannot track ProviderConfig usage")
+		}
+
 		ps := terraform.Setup{}
-		awsCfg, err := getAWSConfig(ctx, c, mg)
+		awsCfg, err := GetAWSConfigWithDefaultRegion(ctx, c, mg, pc)
 		if err != nil {
 			return terraform.Setup{}, errors.Wrap(err, "cannot get aws config")
```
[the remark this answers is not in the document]
erhancagirici: Agreed, but I have another suggestion.
The TF setup caller (in this case upjet connect implementation) already knows that it is calling for a terraform setup and it should be the one wrapping it with `cannot prepare the Terraform setup` (and it already does).

Here, IMHO we should give info about the "step" that TF setup failed, in order to properly distinguish and follow the errors. On top of that I suggest removing the error wrapping inside `GetAWSConfigWithDefaultRegion` implementation and let it return the downstream `GetAWSConfigWithoutTracking` error, because I realized it does not provide any extra context.

internal/clients/aws.go
erhancagirici: agreed, I can leave a comment here.

The thing is, TF config attribute `skip_metadata_api_check` is implemented as a "nullable boolean" at TF side, because the corresponding AWS config `EC2MetadataServiceEnableState` is a 3-state enum. 
- nil ->`ClientDefaultEnableState`, 
- false ->`ClientEnabled`, 
- true ->`ClientDisabled`

In the XP provider config API , `skip_metadata_api_check` is a regular bool, not a bool pointer. Therefore we cannot capture the 3-state. Would it be OK to change the `skip_metadata_api_check` to a bool pointer?

ulucinar: Hi @erhancagirici,
Thank you for the explanation. Because the `skip_metadata_api_check` is an optional field, we should change it to a `bool`  pointer per Kubernetes API conventions (and we should switch to the camel case naming convention). But I suggest we leave this out of scope for this PR, as long as we preserve the old behavior regarding this metadata API check skipping switch before these changes.

We can revisit the ProviderConfig API in future iterations.

internal/clients/aws.go
erhancagirici: Yes, double-checked this and the tags are shown up in console and changes are reflected correctly.

ulucinar:
The `publish-artifacts` job is expected to fail due to the recent Upbound SSO changes. We will need to fix the credentials related to that job.

PR #1233, opened 2024-03-22
Remove ARN Parameter from Firehose DeliveryStream Resource

blakeromano:
This removes the ARN parameter from the Firehose DeliveryStream Resource. After a conversation with @mbbush in this thread https://crossplane.slack.com/archives/C05E0UE46S2/p1711042281273589 this is an attribute we should move to status. I also took some time to optimize the way arn's for other resources were referenced

I have:

Uptest: https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/8693667532


turkenf:
/test-examples="examples/firehose/v1beta1/deliverystream.yaml"

turkenf (approved):
Many thanks for the catch and effort here @blakeromano, LGTM.

`FYI:` @ulucinar, as we discussed, I can create the resource by setting the arn parameter with Terraform Provider without encountering any problems. In this case, the issue seems to be caused by the underlying provider.

PR #1241, opened 2024-03-27
Count external API calls

mergenci:

This PR introduces three AWS API call counters:

1. An AWS SDK v2 [middleware](https://aws.github.io/aws-sdk-go-v2/docs/middleware/) to count authentication calls,
2. An AWS SDK v2 middleware to count resource API calls,
3. An AWS SDK v1 [session handler](https://docs.aws.amazon.com/sdk-for-go/api/aws/session/#hdr-Adding_Handlers) to count resource API calls.

API calls that couldn't be completed because of a connection error are not counted. API calls that return API errors (or no errors) are counted. There are no comprehensive AWS documentation on request rate limits, but here are two resources on the topic, for reference:

1. [Request throttling for the Amazon EC2 API](https://docs.aws.amazon.com/ec2/latest/devguide/ec2-api-throttling.html)
2. [Managing and monitoring API throttling in your workloads](https://aws.amazon.com/blogs/mt/managing-monitoring-api-throttling-in-workloads/)

This PR also removes unsafe pointer operations, as described in https://github.com/upbound/terraform-provider-aws/pull/196.

### Alternatives considered

During the design phase, we investigated whether implementing an [http.RoundTripper](https://pkg.go.dev/net/http#RoundTripper) would be a good solution. Ideally, we would have a common implementation for AWS SDK v1 and v2, since both methods use an [http.Client](https://pkg.go.dev/net/http#Client) under the hood. RoundTripper implementation proved to be infeasible, because of the following reasons:

1. Plugging in a RoundTripper to the client returned by [AWSClient.HTTPClient()](https://github.com/upbound/terraform-provider-aws/blob/af5cc8a3163a2eff74045dc930e83f2fd2bb33b0/internal/conns/awsclient.go#L114) worked for AWS SDK v1 calls, but not for AWS SDK v2 calls.
2. AWS SDK v1 doesn't store service ID (EC2, IAM, etc.) and operation name (DescribeVPCs, etc.) in the [request context](https://pkg.go.dev/net/http#Request.Context), like AWS SDK v2 does. Therefore, we wouldn't be able to label v1 calls by service ID and operation name.
 

### Checklist

I have:

I couldn't run `make reviewable`, because my local terraform setup is broken.

I've tested the code manually using the following resource configuration below, which contains resources that use AWS SDK v1 and v2, as of this writing. Because Upjet comes with Prometheus client, Upjet-based providers serve their metrics at `:8080/metrics`, by default. Here's a sample excerpt after applying the resource configuration:

```
# HELP upjet_resource_external_api_calls The number of external API calls.
# TYPE upjet_resource_external_api_calls counter
upjet_resource_external_api_calls{service="EC2",service_operation="AuthorizeSecurityGroupIngress"} 1
upjet_resource_external_api_calls{service="EC2",service_operation="CreateSecurityGroup"} 1
upjet_resource_external_api_calls{service="EC2",service_operation="CreateTags"} 1
upjet_resource_external_api_calls{service="EC2",service_operation="CreateVpc"} 1
upjet_resource_external_api_calls{service="EC2",service_operation="DescribeNetworkAcls"} 3
upjet_resource_external_api_calls{service="EC2",service_operation="DescribeRouteTables"} 3
upjet_resource_external_api_calls{service="EC2",service_operation="DescribeSecurityGroupRules"} 5
upjet_resource_external_api_calls{service="EC2",service_operation="DescribeSecurityGroups"} 11
upjet_resource_external_api_calls{service="EC2",service_operation="DescribeVpcAttribute"} 9
upjet_resource_external_api_calls{service="EC2",service_operation="DescribeVpcs"} 4
upjet_resource_external_api_calls{service="EC2",service_operation="RevokeSecurityGroupEgress"} 2
upjet_resource_external_api_calls{service="STS",service_operation="GetCallerIdentity"} 1
```

I manually cross-checked reported counts with the calls reported by [CloudTrail Event History](https://us-west-1.console.aws.amazon.com/cloudtrailv2/home?region=us-west-1#/events?ReadOnly=false). Note that CloudTrail Event History may take up to a few minutes to show latest calls.

To test connection errors, I put breakpoints in the code, shut down my Internet connection upon hitting the breakpoint, and then resumed execution. To test API errors, I tried to delete a VPC that has a Security Group configured.

#### Resource Configuration

```yaml
apiVersion: ec2.aws.upbound.io/v1beta1
kind: VPC
metadata:
  annotations:
    meta.upbound.io/example-id: ec2/v1beta1/securitygroupingressrule
  name: test-pr-1241-vpc
  labels:
    testing.upbound.io/example-name: test-pr-1241-vpc
spec:
  forProvider:
    region: us-west-1
    cidrBlock: 172.16.0.0/16
    tags:
      Name: TestPr1241VPC

---
apiVersion: ec2.aws.upbound.io/v1beta1
kind: SecurityGroup
metadata:
  annotations:
    meta.upbound.io/example-id: ec2/v1beta1/securitygroupingressrule
  name: test-pr-1241-securitygroup
  labels:
    testing.upbound.io/example-name: test-pr-1241-securitygroup
spec:
  forProvider:
    region: us-west-1
    vpcIdSelector:
      matchLabels:
        testing.upbound.io/example-name: test-pr-1241-vpc

---
apiVersion: ec2.aws.upbound.io/v1beta1
kind: SecurityGroupIngressRule
metadata:
  name: test-pr-1241-securitygroupingressrule
spec:
  forProvider:
    cidrIpv4: 10.0.0.0/8
    fromPort: 8080
    ipProtocol: tcp
    region: us-west-1
    securityGroupIdRef:
      name: test-pr-1241-securitygroup
    toPort: 8081
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,8 @@ go 1.21
 
 require (
 	dario.cat/mergo v1.0.0
-	github.com/aws/aws-sdk-go-v2 v1.24.0
+	github.com/aws/aws-sdk-go v1.49.2
+	github.com/aws/aws-sdk-go-v2 v1.24.1
 	github.com/aws/aws-sdk-go-v2/config v1.26.1
 	github.com/aws/aws-sdk-go-v2/credentials v1.16.12
 	github.com/aws/aws-sdk-go-v2/feature/ec2/imds v1.14.10
@@ -17,7 +18,7 @@ require (
 	github.com/aws/smithy-go v1.19.0
 	github.com/crossplane/crossplane-runtime v1.16.0-rc.1.0.20240213134610-7fcb8c5cad6f
 	github.com/crossplane/crossplane-tools v0.0.0-20230925130601-628280f8bf79
-	github.com/crossplane/upjet v1.3.0-rc.0.0.20240314162745-2ef7077f6d16
+	github.com/crossplane/upjet v1.3.0-rc.0.0.20240328123350-4c67d8ebd380
 	github.com/go-ini/ini v1.46.0
 	github.com/google/go-cmp v0.6.0
 	github.com/hashicorp/terraform-json v0.18.0
@@ -44,11 +45,10 @@ require (
 	github.com/antchfx/htmlquery v1.2.4 // indirect
 	github.com/antchfx/xpath v1.2.0 // indirect
 	github.com/apparentlymart/go-textseg/v15 v15.0.0 // indirect
-	github.com/aws/aws-sdk-go v1.49.2 // indirect
 	github.com/aws/aws-sdk-go-v2/aws/protocol/eventstream v1.5.4 // indirect
 	github.com/aws/aws-sdk-go-v2/feature/s3/manager v1.15.7 // indirect
-	github.com/aws/aws-sdk-go-v2/internal/configsources v1.2.9 // indirect
-	github.com/aws/aws-sdk-go-v2/internal/endpoints/v2 v2.5.9 // indirect
+	github.com/aws/aws-sdk-go-v2/internal/configsources v1.2.10 // indirect
+	github.com/aws/aws-sdk-go-v2/internal/endpoints/v2 v2.5.10 // indirect
 	github.com/aws/aws-sdk-go-v2/internal/ini v1.7.2 // indirect
 	github.com/aws/aws-sdk-go-v2/internal/v4a v1.2.9 // indirect
 	github.com/aws/aws-sdk-go-v2/service/accessanalyzer v1.26.5 // indirect
@@ -140,6 +140,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sso v1.18.5 // indirect
 	github.com/aws/aws-sdk-go-v2/service/ssoadmin v1.23.5 // indirect
 	github.com/aws/aws-sdk-go-v2/service/ssooidc v1.21.5 // indirect
+	github.com/aws/aws-sdk-go-v2/service/support v1.19.6 // indirect
 	github.com/aws/aws-sdk-go-v2/service/swf v1.20.5 // indirect
 	github.com/aws/aws-sdk-go-v2/service/timestreamwrite v1.23.6 // indirect
 	github.com/aws/aws-sdk-go-v2/service/transcribe v1.34.5 // indirect
@@ -277,6 +278,6 @@ require (
 
 replace golang.org/x/exp => golang.org/x/exp v0.0.0-20231006140011-7918f672742d
 
-replace github.com/hashicorp/terraform-provider-aws => github.com/upbound/terraform-provider-aws v0.0.0-20240129145938-c69f68a59916
+replace github.com/hashicorp/terraform-provider-aws => github.com/upbound/terraform-provider-aws v0.0.0-20240328111213-f2f0fdd63866
 
 replace github.com/hashicorp/terraform-plugin-log => github.com/gdavison/terraform-plugin-log v0.0.0-20230928191232-6c653d8ef8fb
--- a/internal/clients/aws.go
+++ b/internal/clients/aws.go
@@ -6,12 +6,14 @@ package clients
 
 import (
 	"context"
-	"reflect"
-	"unsafe"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	awsmiddleware "github.com/aws/aws-sdk-go-v2/aws/middleware"
 	"github.com/aws/aws-sdk-go-v2/feature/ec2/imds"
+	awsrequest "github.com/aws/aws-sdk-go/aws/request"
+	"github.com/aws/smithy-go/middleware"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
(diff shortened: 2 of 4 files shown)


internal/clients/aws.go
```diff
@@ -111,6 +113,37 @@ func (m *metaOnlyPrimary) Meta() any {
 	return m.meta
 }
 
+// withExternalAPICallCounter configures an AWS SDK v2 stack (client)
+// with an API call counter.
+func withExternalAPICallCounter(stack *middleware.Stack) error {
+	externalAPICallCounterMiddleware := middleware.DeserializeMiddlewareFunc("externalAPICallCounter",
+		func(ctx context.Context, input middleware.DeserializeInput, next middleware.DeserializeHandler) (middleware.DeserializeOutput, middleware.Metadata, error) {
+			serviceID := awsmiddleware.GetServiceID(ctx)
```
ulucinar, line 121: Did we handle the inconsistent service names between the AWS SDK v1 & v2? I feel it would be less confusing if we can emit the same name regardless of the version of the SDK the resource uses.

internal/clients/aws.go
```diff
@@ -111,6 +113,37 @@ func (m *metaOnlyPrimary) Meta() any {
...
+			operationName := awsmiddleware.GetOperationName(ctx)
+
+			output, metadata, err := next.HandleDeserialize(ctx, input)
```
ulucinar, line 124: Could you please explain explicitly in a comment that `next.HandleDeserialize` actually invokes the rest of the chain including the round-tripper and reports back the result?

internal/clients/aws.go
```diff
@@ -111,6 +113,37 @@ func (m *metaOnlyPrimary) Meta() any {
 	return m.meta
 }
 
+// withExternalAPICallCounter configures an AWS SDK v2 stack (client)
+// with an API call counter.
+func withExternalAPICallCounter(stack *middleware.Stack) error {
```
ulucinar, line 118: Maybe it will help the future maintainers to explain how this middleware works, i.e., AWS Go SDK v2 makes the service name and the operation name available in the context and we install a middleware to extract them from the context and increment the upjet metric with the appropriate label values.

internal/clients/aws.go
```diff
@@ -111,6 +113,37 @@ func (m *metaOnlyPrimary) Meta() any {
...
+
+			output, metadata, err := next.HandleDeserialize(ctx, input)
+			if err == nil {
+				metrics.ExternalAPICalls.WithLabelValues(serviceID, operationName).Inc()
+			}
+			return output, metadata, err
+		},
+	)
+
+	// We register the call counter to the end of the deserialization
```
ulucinar, line 132: Thank you for the detailed explanation here.

internal/clients/aws.go
```diff
@@ -162,5 +194,20 @@ func configureNoForkAWSClient(ctx context.Context, ps *terraform.Setup, config *
 	ps.Meta = tfAwsConnsClient
 	fwProvider := xpprovider.GetFrameworkProviderWithMeta(&metaOnlyPrimary{meta: tfAwsConnsClient})
 	ps.FrameworkProvider = fwProvider
+
+	// Register AWS SDK v1 call counter. To learn how SDK v1 session
+	// handler phases map to SDK v2 middleware stack steps, see:
+	// https://aws.github.io/aws-sdk-go-v2/docs/migrating/#handler-phases
+	tfAwsConnsClient.Session.Handlers.Send.PushBack(func(r *awsrequest.Request) {
```
ulucinar, line 201: Maybe we can also mention that the middleware context of SDK v1 does not contain the service & operation names  so we had to use this approach.

ulucinar (commented):
Thank you very much @mergenci for working on this important observability topic. Left some comments for you to consider. Also would you like to briefly discuss the wrapped `http.RoundTripper` approach in the PR description so that we record the challenges associated with that alternative?

internal/clients/aws.go
mergenci: We did. Inconsistency in naming was due to the fact that SDK v1 had `ServiceName`, which we used, in addition to `ServiceID`, which we should have used. Now that both SDK v1 and v2 use `ServiceID`, they all count towards the same metric.

go.mod
```diff
@@ -8,7 +8,8 @@ go 1.21
 
 require (
 	dario.cat/mergo v1.0.0
-	github.com/aws/aws-sdk-go-v2 v1.24.0
+	github.com/aws/aws-sdk-go v1.49.2
+	github.com/aws/aws-sdk-go-v2 v1.24.1
```
ulucinar, line 12: These bumps are coming from the native provider's upbound fork dependency versions.

ulucinar (approved):
Thank you @mergenci, lgtm.

ulucinar:

internal/clients/aws.go
```diff
@@ -6,12 +6,14 @@ package clients
 
 import (
 	"context"
-	"reflect"
-	"unsafe"
```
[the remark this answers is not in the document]
mergenci: 😎🎉

PR #1245, opened 2024-03-28
Use newlines in config/generated.lst to prevent git conflicts

mbbush:

Currently, every pull request that adds a new resource kind modifies the one and only line of `config/generated.lst`, which means that git finds a conflict between any two such pull requests. 

This change makes such PRs no longer be conflicting, by introducing newlines. 

The only potential downside I can see here is that the conflict was serving as a sign that there were some additional changes that _might_ impact the codegen, and you should rebase and rerun `make generate`. But they're a poor way to indicate that. If we really want to ensure that, we can enforce a linear history, but I think simply running CI on main is probably good enough. 

This will also make it much easier to maintain that linear history, since you'll be able to git merge without having to rerun codegen, and the ci `check-diff` job will ensure you didn't miss anything.

I have:


turkenf (approved):
Thank you for taking care of this @mbbush, LGTM.

mbbush:
Thanks @turkenf. I rebased and resolved conflicts hopefully for the last time. I'll be AFK by the time the lint job finishes. Can you take care of clicking merge?

Issue #1252, opened 2024-04-03
[Bug]: 1.3.0 aws-provider breaks pod identity credential resolution


david-kirby:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Affected Resource(s)

* aws.upbound.io/v1beta1
* pkg.crossplane.io/v1
* sqs.aws.upbound.io/v1beta1 - queue

### Resource MRs required to reproduce the bug

In short, the below manifests:
1. setup an aws-sqs provider that gets patched to have a service account named `crossplane`
2. sets up the aws-provider config as `prod` and to use IRSA for credentials (IMPORTANT NOTE: This is working with a Pod Association configuration in EKS. The IAM role is **not** configured for IRSA and instead it's configured for Pod Identity)
3. deploys a simple sqs queue
```yaml
apiVersion: sqs.aws.upbound.io/v1beta1
kind: Queue
metadata:
  name: my-queue
spec:
  forProvider:
    name: my-queue
    region: us-east-1
  providerConfigRef:
    name: prod
---
apiVersion: aws.upbound.io/v1beta1
kind: ProviderConfig
metadata:
  name: prod
spec:
  credentials:
    source: IRSA
---
apiVersion: pkg.crossplane.io/v1
kind: Provider
metadata:
  name: provider-aws-sqs
spec:
  package: xpkg.upbound.io/upbound/provider-aws-sqs:v1.2.1
  runtimeConfigRef:
    name: patch-service-account
---
apiVersion: pkg.crossplane.io/v1beta1
kind: DeploymentRuntimeConfig
metadata:
  name: patch-service-account
spec:
  deploymentTemplate:
    spec:
      selector: {}
      template:
        spec:
          serviceAccountName: crossplane
          containers: []

```

### Steps to Reproduce

I had the above manifests deployed, but upon updating the sqs provider to 1.3.0 and deploying a new queue or deleting/recreating the example, the queue cannot be created.

### What happened?

Using version 1.2.1, the queue is created. After upgrading to 1.3.0, the resource does not get created. I tested this with other resources as well (sns, eks:clusterauth) and encountered the same error message whenever the provider version being used was 1.3.0

### Relevant Error Output Snippet

```shell
message: 'connect failed: cannot initialize the Terraform plugin SDK async external
        client: cannot get terraform setup: cache manager failure: cannot calculate
        the hash for the credentials file: token file name cannot be empty'
```
```


### Crossplane Version

1.15.1

[...]

PR #1235, opened 2024-03-22
Cache AWS Config's CredentialsProvider to reduce STS calls

erhancagirici:
### Description of your changes

Fixes #997 
Introduces a global credential cache to reduce AWS STS calls.
Only IRSA credentials are cached.

The new provider cache is a two-layer hierarchical cache. L1 cache is an AWS SDK Go `aws.CredentialsCache` and the L2 cache is for caching the `aws.CredentialsCache`s. The cache key for the L2 cache is derived from the well known IRSA authentication parameters as well as the contents of the OIDC ID token file. The new cache also caches the AWS account ID for a given IRSA configuration and replaces the identity cache for IRSA configurations.

Background:

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable` to ensure this PR is ready for review.
- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.

### How has this code been tested

Tested manually with provider configs with:
- IRSA auth configuration
- IRSA + RoleChain configuration
- WebIdentity + RoleChain configuration [no credential caching with this PR]
- @turkenf has also validated the provider package `index.docker.io/ulucinar/provider-aws-ec2:v1.3.0-0fbbf02b3656352c729396851646d12ef80a1496` for `Upbound` authentication on Upbound Cloud.
- A static (long term) credential configuration (with authentication type `Secret`) has succeeded here: https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/8468707015

Two experiments were done using 4 managed resources (MRs) with a plain IRSA configuration and an IRSA configuration with an assume role chain of length two with the following `ProviderConfig.aws`:
```yaml
apiVersion: aws.upbound.io/v1beta1
kind: ProviderConfig
metadata:
  name: default
spec:
  assumeRoleChain:
  - roleARN: arn:aws:iam::<account ID>:role/alper-rc-1
  - roleARN: arn:aws:iam::<account ID>:role/alper-rc-2
  credentials:
    source: IRSA
```

During these experiments, we forced frequent reconciliations of the MRs (every 3 seconds) in constant update loops and
we also observed the AWS CloudTrail event history for an extended period of time. Here are the relevant events from CloudTrail:
image

[...]

PR #1254, opened 2024-04-03
feat(pc): add providerconfig example for aws eks pod identity feature

### LINKED ISSUE #1249: Support Pod Identity Controller associations for provider IAM permissions

I’d like to utilise pod identity controller associations to give my AWS providers the correct IAM privileges, instead of IRSA.

I’m not sure if it’s possible or not as the documentation doesn’t seem to indicate it is, if it is I’d love to see a sample ProviderConfig object for this authentication and authorisation mechanism.

haarchri:
add providerconfig example for aws eks pod identity feature (https://aws.amazon.com/blogs/containers/amazon-eks-pod-identity-a-new-way-for-applications-on-eks-to-obtain-iam-credentials/)  and additional hints for IAM Role and OpenIDConnectProvider
realted to this convo: https://github.com/crossplane-contrib/provider-upjet-aws/issues/1252

Fixes #1249 #1252

I have:

### DIFF
--- a/examples/providerconfig/v1beta1/pod-identity.yaml
+++ b/examples/providerconfig/v1beta1/pod-identity.yaml
@@ -23,7 +23,6 @@ metadata:
 spec:
   forProvider:
     clientIdList:
-    - sts.amazonaws.com
     - pods.eks.amazonaws.com
     thumbprintList:
     - 9e99a48a9960b14926bb7f3b02e22da2b0ab7280
@@ -33,19 +32,14 @@ apiVersion: iam.aws.upbound.io/v1beta1
 kind: Role
 metadata:
   name: pod-identity-role
+  labels:
+    testing.upbound.io/example-name: pod-identity
 spec:
   forProvider:
     assumeRolePolicy: |
       {
         "Version": "2012-10-17",
         "Statement": [
-          {
-            "Effect": "Allow",
-            "Principal": {
-              "Federated": "arn:aws:iam::<AWS-ACCOUNT-ID>:oidc-provider/oidc.eks.<REGION>.amazonaws.com/id/<UNIQUE_ID>"
-            },
-            "Action": "sts:AssumeRoleWithWebIdentity"
-          },
           {
             "Effect": "Allow",
             "Principal": {
--- a/examples/providerconfig/v1beta1/pod-identity.yaml
+++ b/examples/providerconfig/v1beta1/pod-identity.yaml
@@ -1,0 +1,60 @@
+# SPDX-FileCopyrightText: 2024 The Crossplane Authors <https://crossplane.io>
+#
+# SPDX-License-Identifier: CC0-1.0
+
+apiVersion: aws.upbound.io/v1beta1
+kind: ProviderConfig
+metadata:
+  name: pod-identity
+spec:
+  credentials:
+    source: WebIdentity
+    webIdentity:
+      roleARN: arn:aws:iam::123456789012:role/pod-identity-role
+      tokenConfig:
+        fs:
+          path: /var/run/secrets/pods.eks.amazonaws.com/serviceaccount/eks-pod-identity-token
+        source: Filesystem
+---
+apiVersion: iam.aws.upbound.io/v1beta1
+kind: OpenIDConnectProvider
+metadata:
+  name: pod-identity
+spec:
+  forProvider:
+    clientIdList:
+    - sts.amazonaws.com
+    - pods.eks.amazonaws.com
+    thumbprintList:
+    - 9e99a48a9960b14926bb7f3b02e22da2b0ab7280
+    url: oidc.eks.<REGION>.amazonaws.com/id/<UNIQUE_ID>
+---
+apiVersion: iam.aws.upbound.io/v1beta1
+kind: Role
+metadata:
+  name: pod-identity-role
+spec:
+  forProvider:
+    assumeRolePolicy: |
+      {
+        "Version": "2012-10-17",
+        "Statement": [
+          {
+            "Effect": "Allow",
+            "Principal": {
+              "Federated": "arn:aws:iam::<AWS-ACCOUNT-ID>:oidc-provider/oidc.eks.<REGION>.amazonaws.com/id/<UNIQUE_ID>"
+            },
+            "Action": "sts:AssumeRoleWithWebIdentity"
+          },
+          {
+            "Effect": "Allow",
+            "Principal": {
+              "Service": "pods.eks.amazonaws.com"
+            },
+            "Action": [
+              "sts:AssumeRole",
+              "sts:TagSession"
+            ]
+          }
+        ]
+      }


examples/providerconfig/v1beta1/pod-identity.yaml
```diff
@@ -0,0 +1,60 @@
...
+metadata:
+  name: pod-identity
+spec:
+  credentials:
+    source: WebIdentity
+    webIdentity:
+      roleARN: arn:aws:iam::123456789012:role/pod-identity-role
+      tokenConfig:
+        fs:
+          path: /var/run/secrets/pods.eks.amazonaws.com/serviceaccount/eks-pod-identity-token
```
blakeromano, line 16: maybe this is a future improvement in crossplane but it would be nice if we could have this be easier to implement without having to hard code path 🤔

examples/providerconfig/v1beta1/pod-identity.yaml
```diff
@@ -0,0 +1,60 @@
...
+    assumeRolePolicy: |
+      {
+        "Version": "2012-10-17",
+        "Statement": [
+          {
+            "Effect": "Allow",
+            "Principal": {
+              "Federated": "arn:aws:iam::<AWS-ACCOUNT-ID>:oidc-provider/oidc.eks.<REGION>.amazonaws.com/id/<UNIQUE_ID>"
+            },
+            "Action": "sts:AssumeRoleWithWebIdentity"
```
mbbush, line 47: Why does this trust policy allow both `AssumeRole` and `AssumeRoleWithWebIdentity`? This implies to me that there are two different ways to use this role, which is not what I'd expect. I thought part of the advantage of using PodIdentity was not needing to deal with OIDC providers any more.

examples/providerconfig/v1beta1/pod-identity.yaml
```diff
@@ -0,0 +1,60 @@
...
+          },
+          {
+            "Effect": "Allow",
+            "Principal": {
+              "Service": "pods.eks.amazonaws.com"
+            },
+            "Action": [
+              "sts:AssumeRole",
+              "sts:TagSession"
+            ]
```
mbbush, line 57: Why are there no conditions in the trust policy? If we're creating an example for others to follow, we should demonstrate best practices, like ensuring that only the specified pods can assume the role. As I read it, this trust policy allows any pod in any EKS cluster in this aws account to assume the role.

mbbush (commented):
I haven't used the PodIdentity feature in EKS yet, so I may just be misinformed about how this works, but there are a few things I find surprising here, that I think should at least be better documented, if not corrected.

I also think that this example probably needs to be more than just a single manifest.  Including both managed resources that must exist before the provider config will work, and the provider config used to create those resources, in the same manifest, creates a circular dependency.

examples/providerconfig/v1beta1/pod-identity.yaml
mbbush: I'm surprised that it's necessary to set this explicitly, and that the SDK doesn't "Just Work" when instantiated in a pod that's assigned a PodIdentity. Maybe what we really need is to add an explicit `PodIdentity` to the list of supported auth types?

examples/providerconfig/v1beta1/pod-identity.yaml
haarchri: we can drop `AssumeRoleWithWebIdentity` as discussed today in slack

examples/providerconfig/v1beta1/pod-identity.yaml
haarchri: today we need to set it explicit - because our GO SDK is not up2date with this feature

blakeromano: Is there an issue to track that?

examples/providerconfig/v1beta1/pod-identity.yaml
haarchri: we don't need any other conditions because you have a `pod-identity-association` which doing the association between pod identity and this role

haarchri:
```
apiVersion: eks.aws.upbound.io/v1beta1
kind: PodIdentityAssociation
metadata:
  name: example
spec:
  forProvider:
    clusterNameSelector:
      matchLabels:
        testing.upbound.io/example-name: example-cluster
    namespace: crossplane-system
    region: eu-central-1
    roleArnSelector:
      matchLabels:
        testing.upbound.io/example-name: pod-identity
    serviceAccount: provider-aws-xxxxxxxx
```

wesleyorama2:
This helped me greatly with getting this up and running using pod identities. Please be sure to also include the findings in:
https://github.com/crossplane-contrib/provider-upjet-aws/issues/1252 in any documentation as I did have to reference it for the trust policy changes that are needed for this.

aiell0:
Hey there! So I recently dug into Pod Identity and Crossplane and was able to get the following configuration to work:

```
apiVersion: pkg.crossplane.io/v1beta1
kind: DeploymentRuntimeConfig
metadata:
  name: patch-service-account
spec:
  deploymentTemplate:
    spec:
      selector: {}
      template:
        spec:
          serviceAccountName: crossplane
          containers: []
---
kind: ProviderConfig
metadata:
  name: prod
spec:
  credentials:
    source: IRSA
---
apiVersion: pkg.crossplane.io/v1
kind: Provider
metadata:
  name: provider-aws-s3
spec:
  package: xpkg.upbound.io/upbound/provider-aws-s3:v1.1.0
  runtimeConfigRef:
    name: patch-service-account
---
apiVersion: s3.aws.upbound.io/v1beta1
kind: Bucket
metadata:
  generateName: crossplane-bucket-
spec:
  forProvider:
    region: us-east-1
  providerConfigRef:
    name: prod
```

The `crossplane` service account is linked to a pod identity with a role that has Admin privileges (I was just trying to get this to work) and the following trust policy:
```
apiVersion: s3.aws.upbound.io/v1beta1
kind: Bucket
metadata:
  generateName: crossplane-bucket-
spec:
  forProvider:
    region: us-east-1
  providerConfigRef:
    name: prod
```

I only bring this up because it is a lot different from the example provided here in that I did not have to use a web identity at all. Is there something I am missing?

examples/providerconfig/v1beta1/pod-identity.yaml
danielloader: Quick question, if not a clarification:

IRSAv1 had a huge pain point in needing to include the roleARN that you're going to annotate service accounts with, so that the OIDC trust loop worked.

With IRSAv2 (Pod Identity Controller) I can configure all the AWS IAM/EKS components out of cluster, and as long as a pod starts up in the nominated namespace, and with the service account I picked, it should inherit the permissions at start via the injected token provided by the pod access controller.

Given this; if I have to mangle my helm to be coupled to the IAM ARN, I'm not sure I'm saving any hassle here - I want to just add a pod identity out of band (outside EKS, not involving Crossplane, since the host cluster is managed with terraform) and have Crossplane start up and use the IAM permissions granted to it.

Am I misunderstanding things here?

It's a relatively new concept but I've had success with this decoupled model with External Secrets controller for example.

Additionally as pointed out by @mbbush if we need to configure this, rather than letting it fall through the SDK default authentication methods a lot of the benefit of "it just works" that; the pod identity controller, the associated admission mutation webhook and the tokens just being in the expected place... start to lose some value.

danielloader:
@aiell0 interestingly your example works for me with aws upbound provider v1.1.0, but if I bump the provider to v1.4.0 it falls apart. 

The error is: `cannot initialize the Terraform plugin SDK async external client: cannot get terraform setup: cache manager failure: cannot calculate the hash for the credentials file: token file name cannot be empty`

I think the issue is likely covered in the v1.3.0 release: https://github.com/crossplane-contrib/provider-upjet-aws/releases/tag/v1.3.0

> This release also introduces a credential cache for IRSA authentication, which greatly reduces the number of AWS STS calls the provider makes. This cache is currently only employed for IRSA configurations. Please refer to the description https://github.com/crossplane-contrib/provider-upjet-aws/pull/1235#issue-2202869791 for the results of some experiments and the observed improvements in those experiments.

Can confirm 1.1.0, 1.2.1 works and 1.3.0, 1.4.0 don't.

Been banging my head on wall trying to understand why your example applied to the cluster worked a treat, and my own in argocd didn't work at all.

Given this, either the change is a regression, or the IRSA mode was a fluke (seems like it was) and the cache breaks it when you use the Pod Identity token, which is in a different format to the IRSA one.

We might need a dedicated source for `PodIdentity`.

haarchri:
let's double check the Issues with 1.3 and 1.4 and include in the fix an working example without any workaround for hardcoded filesystem files

danielloader:
Thanks for looking at this for us.

truongnht:
@haarchri I am curious if you have update on this issue 🤔

quanlk2511:
Hi @haarchri , guys, do we have progress on this?

haarchri:
please check out this: https://github.com/crossplane-contrib/provider-upjet-aws/pull/1459

PR #1265, opened 2024-04-16
Select spec fields specific to Upbound OIDC

jastang:
Update `auth.yaml` to select specific fields in the spec to support Upbound as an identity provider.

Screenshot of a deploy with a new xpkg:
Screen Shot 2024-04-16 at 9 43 19 AM

Fixes https://github.com/upbound/upbound-frontend/issues/3751

I have:

### DIFF
--- a/package/auth.yaml
+++ b/package/auth.yaml
@@ -52,4 +52,5 @@ sources:
 
     Copy the ARN from the **Summary** section of the role into the input.
   showFields:
-  - spec.credentials.upbound
+  - spec.credentials.upbound.webIdentity.roleARN
+  - spec.credentials.upbound.webIdentity.roleSessionName


ulucinar (approved):
Thank you @jastang. Discussed this issue with @erhancagirici and we will remove the `TokenConfig` API from the `Upbound` authentication scheme once we have the ProviderConfig e2e tests in place. 

If, in the future, we would like to be able to configure the source of the OIDC ID token (e.g., making the location of the token configurable so that we may either read it from a local file (as we do right now), or read it from a secret or similar,  when the auth method is `Upbound`), then we can consider re-adding it to the API and implement the retrieval mechanism on top of that API. Currently, the API is there but the retrieval implementation is missing. And in the managed control planes, we always provision the OIDC token on the filesystem at a well-known path, so a configuration for the token source is currently unnecessary.

The alternative would be just to remove the `TokenConfig` from `Upbound` auth configuration as the API is dummy and I assume the marketplace does not generate a ProviderConfig with the configuration provided by the `TokenConfig` API (as it's currently useless). But this is technically a breaking change. And once we have the ProviderConfig e2e tests in place, it will be a better time to make this change in the ProviderConfig.

ulucinar:
So, my understanding is that it's technically possible, there exist some `ProviderConfig`s in the MCPs using the `Upbound` authentication scheme with the `TokenConfig` configuration set. Is this correct? Before this change, were all the fields under `spec.credentials.upbound` available to configure for the console users and we would provision a `ProviderConfig` using all those configured fields?

Issue #1027, opened 2023-12-15
Lambda Functions keeps updating after few seconds with Diff Detected 


sanketgawade:
### What happened?
We have started using Upblund Family AWS provider with our Crossplane setup. We recently created lambdas using the upbound provider. The initial lambda is fine. After first update the lambda is getting updated by the upbound provider with below logs. Every few seconds the update is happening. Is there a way to stop the sync so frequently , or is this a bug ? 
Crossplane: 1.12.3 
Upbound family aws provider: 0.46.0 
EKS: 1.26

2023-12-15T14:08:44Z    DEBUG   provider-aws    Diff detected   {"uid": "a0d7651d-91d9-4a61-9f60-7e493e69ce8a", "name": "tor-data-feed-loader-mzpp8-h49h2", "gvk": "lambda.aws.upbound.io/v1beta1, Kind=Function", "instanceDiff": "*terraform.InstanceDiff{mu:sync.Mutex{state:0, sema:0x0}, Attributes:map[string]*terraform.ResourceAttrDiff{\"source_code_hash\":*terraform.ResourceAttrDiff{Old:\"39759c7c1e2f08ea8f5ecda48d64f47cfd23874726ebc5024eb10c43c2134ebf\", New:\"83e6074afd44dc85ffb5effde5e222f50815e72e76a15dbf9c726f2a186b5a51\", NewComputed:false, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:false, Sensitive:false, Type:0x0}}, Destroy:false, DestroyDeposed:false, DestroyTainted:false, RawConfig:cty.NilVal, RawState:cty.NilVal, RawPlan:cty.NilVal, Meta:map[string]interface {}(nil)}"}
[...]

PR #1266, opened 2024-04-16
Fix update loop for Function.lambda resource 

turkenf:

This PR fixes the update loop in the `Function.lambda` resource.

Fixes: https://github.com/crossplane-contrib/provider-upjet-aws/issues/1027

I have:

- Create a `Repository.ecr` resource by applying the following:
```
apiVersion: ecr.aws.upbound.io/v1beta1
kind: Repository
metadata:
  name: test-repo
spec:
  forProvider:
    region: us-east-1
    imageTagMutability: "IMMUTABLE"
```
- Push an image to the repository:
   - Retrieve an authentication token and authenticate your Docker client to your registry. Use the AWS CLI: `aws ecr get-login-password --region us-east-1 | docker login --username AWS --password-stdin <account_id>.dkr.ecr.us-east-1.amazonaws.com`
   - Build your Docker image using the following command. For information on building a Docker file from scratch see the instructions [here ](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/docker-basics.html). You can skip this step if your image is already built: `docker build -t test-repo .`
   - After the build completes, tag your image so you can push the image to this repository: `docker tag test-repo:latest <account_id>.dkr.ecr.us-east-1.amazonaws.com/test-repo:latest`
   - Run the following command to push this image to your newly created AWS repository: `docker push <account_id>.dkr.ecr.us-east-1.amazonaws.com/test-repo:latest`
- Create a `Function.lambda` resource by applying the following:
```
apiVersion: lambda.aws.upbound.io/v1beta1
kind: Function
metadata:
  name: example-ft-test
spec:
  forProvider:
    imageUri: <redacted>.dkr.ecr.us-east-1.amazonaws.com/ft-test-repo:latest
    packageType: Image
    region: us-east-1
    roleSelector:
      matchLabels:
        testing.upbound.io/example-name: role
    timeout: 60
``` 
- Once the resource has been created successfully, create and push a different image with a tag using the steps above.
- Change the `imageUri` in the `Function.lambda` resource with the new one.
```
2024-04-16T19:22:57+03:00	DEBUG	provider-aws	External resource is up to date	{"controller": "managed/lambda.aws.upbound.io/v1beta1, kind=function", "request": {"name":"example-ft-test"}, "uid": "786b127b-d342-423f-89b0-9e45c61b1d77", "version": "36662", "external-name": "example-ft-test", "requeue-after": "2024-04-16T19:33:03+03:00"}
```
```
  conditions:
  - lastTransitionTime: "2024-04-16T16:21:44Z"
    reason: Available
    status: "True"
    type: Ready
  - lastTransitionTime: "2024-04-16T16:21:35Z"
    reason: ReconcileSuccess
    status: "True"
    type: Synced
  - lastTransitionTime: "2024-04-16T16:21:42Z"
    reason: Success
    status: "True"
    type: LastAsyncOperation
  - lastTransitionTime: "2024-04-16T16:22:57Z"
    reason: UpToDate
    status: "True"
    type: Test
```

### DIFF
--- a/config/lambda/config.go
+++ b/config/lambda/config.go
@@ -6,6 +6,7 @@ package lambda
 
 import (
 	"github.com/crossplane/upjet/pkg/config"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 
 	"github.com/upbound/provider-aws/config/common"
 )
@@ -61,6 +62,14 @@ func Configure(p *config.Provider) {
 			SelectorFieldName: "SubnetIDSelector",
 		}
 		delete(r.TerraformResource.Schema, "filename")
+
+		r.TerraformCustomDiff = func(diff *terraform.InstanceDiff, _ *terraform.InstanceState, _ *terraform.ResourceConfig) (*terraform.InstanceDiff, error) {
+			if diff != nil && diff.Attributes != nil {
+				delete(diff.Attributes, "source_code_hash")
+				delete(diff.Attributes, "last_modified")
+			}
+			return diff, nil
+		}
 	})
 
 	p.AddResourceConfigurator("aws_lambda_function_event_invoke_config", func(r *config.Resource) {
--- a/config/lambda/config.go
+++ b/config/lambda/config.go
@@ -64,6 +64,10 @@ func Configure(p *config.Provider) {
 		r.LateInitializer = config.LateInitializer{
 			IgnoredFields: []string{"source_code_hash"},
 		}
+		r.MetaResource.ArgumentDocs["source_code_hash"] = "Used to trigger updates. Must be set to " +
+			"a base64 encoded SHA256 hash of the package file specified with either filename or s3_key. " +
+			"If you have specified this field manually, it should be the actual (computed) hash of the " +
+			"underlying lambda function specified in the filename, image_uri, s3_bucket fields."
 	})
 
 	p.AddResourceConfigurator("aws_lambda_function_event_invoke_config", func(r *config.Resource) {
--- a/config/lambda/config.go
+++ b/config/lambda/config.go
@@ -6,7 +6,6 @@ package lambda
 
 import (
 	"github.com/crossplane/upjet/pkg/config"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 
 	"github.com/upbound/provider-aws/config/common"
 )
@@ -62,13 +61,8 @@ func Configure(p *config.Provider) {
 			SelectorFieldName: "SubnetIDSelector",
 		}
 		delete(r.TerraformResource.Schema, "filename")
-
-		r.TerraformCustomDiff = func(diff *terraform.InstanceDiff, _ *terraform.InstanceState, _ *terraform.ResourceConfig) (*terraform.InstanceDiff, error) {
-			if diff != nil && diff.Attributes != nil {
-				delete(diff.Attributes, "source_code_hash")
-				delete(diff.Attributes, "last_modified")
-			}
-			return diff, nil
+		r.LateInitializer = config.LateInitializer{
+			IgnoredFields: []string{"source_code_hash"},
 		}
 	})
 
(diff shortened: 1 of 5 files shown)


config/lambda/config.go
```diff
@@ -61,6 +62,14 @@ func Configure(p *config.Provider) {
 			SelectorFieldName: "SubnetIDSelector",
 		}
 		delete(r.TerraformResource.Schema, "filename")
+
+		r.TerraformCustomDiff = func(diff *terraform.InstanceDiff, _ *terraform.InstanceState, _ *terraform.ResourceConfig) (*terraform.InstanceDiff, error) {
+			if diff != nil && diff.Attributes != nil {
+				delete(diff.Attributes, "source_code_hash")
```
ulucinar, line 68: From the [Terraform docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_function#source_code_hash), looks like this argument is used to trigger updates probably in situations where the image URI does not change but the contents of the image identified by that URI changes. 

So, we had better consider this argument while calculating the diffs. My understanding the problem here is Terraform expects from the users of this API to update the hash accordingly when the image URI (or more precisely when the content is) updated. If this is really the case, what we can do is to make this dependency explicit in the parameter's documentation.

In addition to this, we had better disable late-initialization for this field because if the above explanation is correct, then whenever a user specified an image URI but leaves the hash empty, could it be the case that the provider late-initializes this field with the computed hash, making it a requirement for the user to update the hash if they just want to update the image URI? If this is the case, then disabling late-initialization would relieve users of this API, who just rely on the image URI (and don't want to trigger an update by updating the hash).

config/lambda/config.go
```diff
@@ -61,6 +62,14 @@ func Configure(p *config.Provider) {
...
+				delete(diff.Attributes, "last_modified")
```
ulucinar, line 69: In our generated API, `last_modified` is an [observation field](https://github.com/crossplane-contrib/provider-upjet-aws/blob/d29baaf5c73aacdcc9145990a9e2e19f7be6ef9e/apis/lambda/v1beta1/zz_function_types.go#L302), i.e., an attribute. So it should already be not considered in drift calculation (it's not part of the desired state).

ulucinar (commented):
Thank you @turkenf, left some comments for discussion. If the discussions are correct, then I think this works more or less as intended and to improve the situation, we can disable the late-initialization for the hash field (so that users don't have to update the hash if they don't care about it) because late-initialization forces them to care about the field :)

ulucinar:
@turkenf, thanks for documenting the manual testing steps. We should also consider capturing this know-how in the uptest manifests (probably by executing them via the pre hooks) and automating the testing. We may even consider using an already available test image (I don't know its cost implications tbh).

config/lambda/config.go
blakeromano: I think the usage of this is really for zip based lambda deployments (where contents are in S3). I don't think `source_code_hash` has anything to do with image based deployments. I have at least yet to see any documentation for container based Lambdas where this is referenced. I think maybe what we could do is see if the attribute for `image_uri` is not nil and if it is not nil then delete the `source_code_hash`?

turkenf: @ulucinar, after disabling late-initialization for `source_code_hash`, I do not encounter the update loop (unless the user has set `source_code_hash` manually) and I think it's better to continue this way.

@blakeromano, although I do not know the exact use case, after the tests I have done, I think that the diff for `source_code_hash` should not be deleted. In case the `image_uri` remains the same for an image in use and the `source_code_hash` changes, it is necessary to set the source_code_hash to update the resource.

config/lambda/config.go
```diff
@@ -61,6 +61,9 @@ func Configure(p *config.Provider) {
 			SelectorFieldName: "SubnetIDSelector",
 		}
 		delete(r.TerraformResource.Schema, "filename")
+		r.LateInitializer = config.LateInitializer{
+			IgnoredFields: []string{"source_code_hash"},
```
blakeromano, line 65: The only thing I'd validate here is how this may impact zip based lambdas in S3. I am not sure how setting a late initializer on that may impact this. I think an uptest should alleviate any concerns there though.

mbbush:
I'm very supportive of @ulucinar's suggestion to fix this by disabling late initialization. It seems like that should fix the issues, but I'd also like to try some creative ideas to break it that will hopefully fail, both for zip lambdas and image-based ones.

AWS Lambda requires that its images come from AWS ECR in the same region as the lambda, but they can be from any aws account, as long as permissions are set appropriately. So I think it would work well to have a small, static lambda image hosted in ECR in a different aws account than is used for testing (so it doesn't get cloud-nuke'd), that we could use for testing from any aws account. It's worth checking, but I'd expect the cost to be minimal.

PR #1271, opened 2024-04-20
Upgrade to terraform provider v5.46.0

mbbush:
### Description of your changes
Upgrade to the latest terraform provider version. There have been several complications, discussed below.

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable` to ensure this PR is ready for review.

### Complications
This is integrating several separate pieces, which conflict with each other.

- [ ] AWSClient.Session is now private
  - Since https://github.com/crossplane-contrib/provider-upjet-aws/pull/1241 we're invoking the `.Session` property on `conns.AWSClient`
  - In https://github.com/hashicorp/terraform-provider-aws/commit/b87c156fa3ac1199937815266d6f839b509572f3 the terraform AWS provider made the `.Session` property private (which isn't a breaking change to their public API since it's in `internal`, but we use a forked version of the TF provider that exposes some internal objects to make our no-fork architecture possible). 
  - Thanks to a suggestion from @mergenci I added a getter in my fork of the terraform provider. This appears to be working but I would like help verifying that
  - [ ] metrics are still working properly for both go sdk v1 and v2
- [ ] EKS ClusterAuth
  - In https://github.com/crossplane-contrib/provider-upjet-aws/issues/1248 we observed a regression in the cluster auth generated for an EKS cluster when upgrading to aws go sdk v2 version 1.24.1. We resolved it by pinning the aws go sdk v2 version to 1.24.0.
  - The best explanation I could find for what happened was in https://github.com/eksctl-io/eksctl/pull/7487. `eksctl` resolved it by removing the `RetryMetricsHeader` from the middleware, but then later [reverted that change](https://github.com/eksctl-io/eksctl/pull/7563) after upgrading to aws go sdk v2 version 1.25.0.
  - The new and improved `ClusterAuth` uptest passes on this branch, but I'd still appreciate
  - [ ] Someone with more context on the issue validating that it's not recurring
- [x] EMR Serverless Application type change
  - https://github.com/aws/aws-sdk-go-v2/pull/2458 introduced a breakfix change to emrserverless, changing a type, which causes a compilation error if we upgrade just the aws sdk without also consuming the [corresponding fix](https://github.com/hashicorp/terraform-provider-aws/commit/b90fdc3f66dd11ae79d75545541e112f382ec40c) from the terraform aws provider.
  - Once consuming both pieces, uptest passes :heavy_check_mark: for the emrserverless application

### Potentially incompatible schema changes:
[...]

PR #1273, opened 2024-04-21
Add EKS access entry and policy association resources

### LINKED ISSUE #1098: Request for `aws_eks_access_entry ` resource

Terraform Resource Name: aws_eks_access_entry, aws_eks_access_policy_association

### What is your use case?

Recently AWS EKS launched simplified Kubernetes Access to replace authentication via `aws-auth` configmap.
https://aws.amazon.com/blogs/containers/a-deep-dive-into-simplified-amazon-eks-access-management-controls/

This feature was recently [released](https://github.com/hashicorp/terraform-provider-aws/issues/34982) in Terraform

mbbush:

Add support for api-based auth in EKS, using eks access entries and access policy associations. 

Note that this PR also includes commits from https://github.com/crossplane-contrib/provider-upjet-aws/pull/1271, because this resource requires a terraform provider upgrade. Testing that upgrade should be a separate effort, done on that PR or a similar one (I think @haarchri was interested in working on the upgrade too?). 

Fixes #1098 

I have:

I applied the uptest example manifest locally, and successfully logged in using the specified role (not the role that created the cluster).

I'm planning to update the uptest example manifest to include an update test as well.


johnathan-sq:
Are there any updates on this PR? Keen to see this in main

fernandezcuesta:
I guess the underlying version constraint is gone with 1.5?

haarchri:
Yes

mbbush:
/test-examples="examples/eks/v1beta1/accessentry.yaml"

mbbush:
/test-examples="examples/eks/v1beta1/accessentry.yaml"

package/crds/eks.aws.upbound.io_accesspolicyassociations.yaml
```diff
@@ -0,0 +1,540 @@
...
+                  field in a future release. Currently, both could be set independently and
+                  non-default values would be honored if the feature flag is enabled.
+                  See the design doc for more information: https://github.com/crossplane/crossplane/blob/499895a25d1a1a0ba1604944ef98ac7a1a71f197/design/design-doc-observe-only-resources.md?plain=1#L223
+                enum:
+                - Orphan
+                - Delete
+                type: string
+              forProvider:
+                properties:
+                  accessScope:
```
mbbush, line 77: @ulucinar This is a singleton list ([MaxItems: 1](https://github.com/hashicorp/terraform-provider-aws/blob/f95c2491260f09e014541e0e94368e0f1f4b690a/internal/service/eks/access_policy_association.go#L45-L52)). Do you have a preference about whether to have two api versions or just one (with an embedded object)?

config/externalname.go
```diff
@@ -2989,6 +2994,12 @@ func getPermissionSetId(tfstate map[string]any) (string, error) {
 // IDs that use elements from the parameters in a certain string format.
 // It should be used in cases where all information in the ID is gathered from
 // the spec and not user defined like name. For example, zone_id:vpc_id.
+//
+// TODO: This should set keys as IdentifierFields, because if they're missing observe-only resources won't work.
+// But that would remove them from spec.initProvider, which would be a breaking schema change for existing resources
+// that we don't have a good way to handle yet.
+//
+// For new resources, prefer using FormattedIdentifierFromProviderWithIdentifierFields instead.
```
mbbush, line 3002: @ulucinar I'd love to come up with a better solution for the (many) cases where we have resources with fields that should be identifier fields, but are not. The options I can think of are:
1. Make a way to include them in both `spec.forProvider` and `spec.initProvider`.
2. Remove immutable fields from `spec.initProvider` in a new api version
3. Remove the need for them to be required by allowing the terraform id to be constructed from either the parameters or the external name.

I think the last would probably be the best long-term user experience, but I don't know how hard it is, or if it would cause other problems. 

What do you think?

config/externalname.go
```diff
@@ -3009,6 +3020,19 @@ func FormattedIdentifierFromProvider(separator string, keys ...string) config.Ex
...
 }
 
+// FormattedIdentifierFromProviderWithIdentifierFields is a helper function to construct Terraform
+// IDs that use elements from the parameters in a certain string format.
+// It should be used in cases where all information in the ID is gathered from
+// the spec and not user defined like name. For example, zone_id:vpc_id.
+// This function sets the keys as IdentifierFields, which means that they are always required, even for observe-only
+// resources. Because the id is constructed exclusively from the keys, omitting them (even if the external name
+// annotation is set) leaves the provider unable to find the terraform id to use to observe the resource.
+func FormattedIdentifierFromProviderWithIdentifierFields(separator string, keys ...string) config.ExternalName {
```
mbbush, line 3030: These function names are getting excessively verbose, but I couldn't think of another option that preserved the "dsl-like" semantics of (most of) the external name config file, didn't introduce a schema change, and could handle the `...string` variable arguments. 

If anyone has a better idea I'd love to hear it.

config/eks/config.go
```diff
@@ -110,4 +109,38 @@ func Configure(p *config.Provider) {
 		}
 		r.UseAsync = true
 	})
+	p.AddResourceConfigurator("aws_eks_access_policy_association", func(r *config.Resource) {
+		r.References = config.References{
+			"cluster_name": {
+				TerraformName: "aws_eks_cluster",
+				// Use the terraform id instead of the external name because the external name is set before the cluster
+				// has been created.
+				Extractor: common.PathTerraformIDExtractor,
```
mbbush, line 118: When I was testing this resource, I was surprised to see that with the default external name extractor, the reference to the eks cluster was being resolved even before the MR posted a Ready condition. I thought that the Ready check was part of all the cross-resource references, but when I went looking for the code that implemented it I couldn't find anything. 

I'm not sure if this is something special about the eks cluster resource, something that would be true of all resources with an external name config that has the name-based initializer enabled, something I'm just misremembering, or a sign of a larger problem. 

It seems like it would be especially problematic for those resources which use an external name which is not the same as the metadata.name, but which have an external name config that has the initializer enabled, because then references would resolve first to the metadata.name (in the external name field) and then once the real external name was set, switch to the right value.

turkenf:
/test-examples="examples/eks/v1beta1/accessentry.yaml"

package/crds/eks.aws.upbound.io_accesspolicyassociations.yaml
turkenf: The PR you mentioned has been merged, so, please rebase your PR and add the v1beta1 version of the resource without the singleton list.

config/externalname.go
turkenf: When I examined the import sections in the two resources you added, the parameters used in id can be taken from `spec.forProvider`. In other words, since these are not resources whose identities are assigned by the remote client, I think it is better to use [TemplatedStringAsIdentifier](https://github.com/crossplane/upjet/blob/37c7f4e91d57982f58e496dfb86dd7c2d29a1921/pkg/config/externalname.go#L70) here if it works.

This function also sets the keys as `IdentifierFields`, so extra configuration is unnecessary.

config/externalname.go
```diff
@@ -437,17 +437,22 @@ var TerraformPluginSDKExternalNameConfigs = map[string]config.ExternalName{
...
 	//
-	"aws_eks_cluster": config.NameAsIdentifier,
-	// Imported using the cluster_name and node_group_name separated by a
-	// colon (:): my_cluster:my_node_group
-	"aws_eks_node_group": config.TemplatedStringAsIdentifier("node_group_name", "{{ .parameters.cluster_name }}:{{ .external_name }}"),
 	// my_cluster:my_eks_addon
 	// "aws_eks_addon": config.TemplatedStringAsIdentifier("addon_name", "{{ .parameters.cluster_name }}:{{ .external_name }}"),
 	"aws_eks_addon": FormattedIdentifierFromProvider(":", "cluster_name", "addon_name"),
+	// import EKS access entry using the cluster_name and principal_arn separated by a colon (:).
+	"aws_eks_access_entry": FormattedIdentifierFromProviderWithIdentifierFields(":", "cluster_name", "principal_arn"),
```
turkenf, line 444: As I mentioned [here](https://github.com/crossplane-contrib/provider-upjet-aws/pull/1273#discussion_r1653641930), could you please try the following:

```suggestion
	"aws_eks_access_entry": config.TemplatedStringAsIdentifier("", "{{ .parameters.cluster_name }}:{{ .parameters.principal_arn }}"),
```

config/externalname.go
```diff
@@ -437,17 +437,22 @@ var TerraformPluginSDKExternalNameConfigs = map[string]config.ExternalName{
...
+	// import EKS access entry using the cluster_name principal_arn and policy_arn separated by a (#) which the tf provider docs incorrectly describe as a colon.
+	"aws_eks_access_policy_association": FormattedIdentifierFromProviderWithIdentifierFields("#", "cluster_name", "principal_arn", "policy_arn"),
```
turkenf, line 446:
```suggestion
	"aws_eks_access_policy_association": config.TemplatedStringAsIdentifier("", "{{ .parameters.cluster_name }}#{{ .parameters.principal_arn }}#{{ .parameters.policy_arn }}"),
```

r-nasiri:
any updates on this? this is a very useful feature specially if you have multiple node groups or need to grant access to multiple IAM roles. updating config map is not fun :)

config/externalname.go
mbbush: I'll put it on the sig-upjet agenda doc.

mbbush: https://github.com/crossplane/upjet/issues/422

fhochleitner:
plus one from my side this feature/provider would realy make our lives a lot easier

turkenf:
Hey @mbbush, are you still interested in taking this PR forward?

config/externalname.go
mbbush: That's exactly what I tried as my first external name configuration; see https://github.com/crossplane-contrib/provider-upjet-aws/pull/1273/commits/24a791a60d863aeebc1b88c91b15540fc564d514. The problem with `TemplatedStringAsIdentifier` is that because it has `DisableNameInitializer: false` it initializes the external name to the `metadata.name`, before updating it to use the value from the template. When I was testing, this caused something to not work; I don't remember what it was. 

Your comment about the identifier not really coming from the provider is a good one; that means this is not a good name for this function. Really the essence of what I need is to disable the name initializer. I think the best solution is to add `WithDisabledNameInitializer` to the `ExternalNameFrom` helper functions @ulucinar added in https://github.com/crossplane/upjet/pull/342/commits/47cb5c5b4bba12627007155937ca23bfc468c0ba

mbbush:
/test-examples="examples/eks/v1beta1/accessentry.yaml"

mbbush:
/test-examples="examples/eks/v1beta1/accessentry.yaml"

config/externalname.go
mbbush: The trouble is that I couldn't find any actual examples of usages of the `ExternalNameFrom` helpers, so I'll probably have to fumble around a bit in order to figure out the syntax.

mbbush:
/test-examples="examples/eks/v1beta1/accessentry.yaml"

mbbush:
/test-examples="examples/eks/v1beta1/accessentry.yaml"

config/externalname.go
turkenf: If the problem is just `DisableNameInitializer: false`, we may use [this function](https://github.com/crossplane-contrib/provider-upjet-aws/blob/be19a8afe756d43e5a79105766001bbb1ca300bd/config/externalname.go#L3095)?

mbbush:
@turkenf I've rebased this, updated the schema to use embedded objects, and reran the tests. I renamed the `...FromProvider` external name configs which only actually used values from the parameters to be named `...FromParameters`, and gave the shorter name to the one that does include identifier fields, as that's what I would expect to be more correct for new resources. We can update the existing resource configurations as needed.

turkenf:
> As I mentioned above, the main difference between the current external name config and the TemplatedString one is that this disables the name initializer, which I see as a good thing, since it would only ever set the external name to an incorrect value.

I left a [comment](https://github.com/crossplane-contrib/provider-upjet-aws/pull/1273#discussion_r1695681382) above.

> @turkenf I've rebased this, updated the schema to use embedded objects, and reran the tests. I renamed the ...FromProvider external name configs which only actually used values from the parameters to be named ...FromParameters, and gave the shorter name to the one that does include identifier fields, as that's what I would expect to be more correct for new resources. We can update the existing resource configurations as needed.

I really appreciate your effort @mbbush, but I think the externalname change should be handled in another PR, even if we are just changing the name.

mbbush:
/test-examples="examples/cloudformation/v1beta1/stack.yaml"

mbbush:
/test-examples="examples/appsync/v1beta1/apikey.yaml"

config/externalname.go
mbbush: `TemplatedStringAsIdentifierWithNoName` is only used by two resources, both of which have a provider-generated component to their terraform id which is separated out and stored in the external name. I recall running into some problems with them in the past related to initial creation, which prompted the TODO I put in that function's docs, and I'm re-running their uptest examples to refresh my memory. I'll check back once those tests finish to see what they did.

Do you have a reason for preferring the "Templated" syntax in the external name config over specifying the parameters and the separator? It seems like that just creates more work to reverse-engineer the template string to do things like find identifier fields and look for `{{ .external_name }}`.

mbbush: I think that whatever problem I was observing on initial creation for `TemplatedStringAsIdentifierWithNoName` was only a problem with the cli-forking architecture, as I saw no issues with those resources. I'll remove the TODO comment from that external name config.

turkenf: > Do you have a reason for preferring the "Templated" syntax in the external name config over specifying the parameters and the separator? 

I don't have any special reason or desire to use the templated syntax. I don't find it appropriate to add a new function for consistency reasons if we can configure the resource without any problems with the existing external name functions/configurations.

> I think that whatever problem I was observing on initial creation for TemplatedStringAsIdentifierWithNoName was only a problem with the cli-forking architecture, as I saw no issues with those resources. I'll remove the TODO comment from that external name config.

From your comment, am I to understand that you will try to continue with `TemplatedStringAsIdentifierWithNoName`?

turkenf:
Hey @mbbush, kindly reminder: https://github.com/crossplane-contrib/provider-upjet-aws/pull/1273#discussion_r1696815920

turkenf:
/test-examples="examples/eks/v1beta1/accessentry.yaml"

mbbush:
@turkenf My plan for handling this was actually to remove `TemplatedStringAsIdentifierWithNoName`, as the two resources that use it work just as well (or better) using `TemplatedStringAsProviderDefinedIdentifier`, which has the added benefits of a name that makes sense with what it does, and not setting the external name to an invalid value (the template with an empty string in the `.external_name` part) during creation, as `TemplatedStringAsIdentifierWithNoName` does for templates that contain `{{ .external_name }}`.

turkenf:
/test-examples="examples/eks/v1beta1/accessentry.yaml"

PR #1277, opened 2024-04-21
Move all example manifests in us-west-1 to us-east-2

mbbush:

The us-west-1 AWS region, like most regions, has three availability zones, but *only two are available to new accounts*. This means that depending on which AWS account you're running in, some of the example manifests may reference an az code that doesn't exist in your account (I have an account that only allows us-west-1b and us-west-1c, but I think the upbound account used by github actions only allows us-west-1a and us-west-1b). 

The simplest way to avoid this is to just avoid the us-west-1 region entirely. This PR replaces all instances of "us-west-1" in the `examples` directory with "us-east-2". us-east-2 is a region that is available to all accounts by default, which always has three AZs available, which doesn't have some of the "special" behavior associated with us-east-1, and which is only referenced in a few of the example manifests.

I have:

Before this change, I found the following distribution of regions in the example manifests:
us-west-1: 640 files
us-east-1: 201 files
us-east-2: 20 files
us-west-2: 32 files
eu-west-1: 6 files

After this change, the distribution was
us-west-1: 0 files
us-east-1: 201 files
us-east-2: 659 files
us-west-2: 33 files
eu-west-1: 6 files

The counts are off by one from what you'd expect from a straight replacement because I additionally edited the db instance automated backup replication manifest, which had previously used both us-east-2 and us-west-1, to use us-east-2 and us-west-2.

### Uptest runs:
:heavy_check_mark:  db instance automated backup replication: https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/8774463999

:heavy_check_mark:  KMS replica external key: https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/8774934667/job/24076679861 

My last commit only updated the readme, so most CI was skipped, but it ran successfully on the penultimate commit at https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/8774918829


mbbush:
/test-examples="examples/rds/v1beta1/dbinstanceautomatedbackupsreplication.yaml"

mbbush:
/test-examples="examples/kms/v1beta1/replicaexternalkey.yaml"

README.md
```diff
@@ -66,6 +66,10 @@ Use `monolith` instead of `ec2` to build and publish the monolithic provider.
 
 Follow the guide [here](https://github.com/crossplane/upjet/blob/v0.10.0/docs/add-new-resource-short.md).
 
+When creating new manifests in the `examples/` directory, please using the `us-west-1` region, as its inconsistent number
```
mbbush, line 69:
```suggestion
When creating new manifests in the `examples/` directory, please avoid using the `us-west-1` region, as its inconsistent number
```

turkenf:
Thank you for your effort and interest in this PR @mbbush, as far as I can see here, we have changed the region in many of the existing resource examples. My only doubt here is the possibility of uptestable resources failing due to the region change. So it would be great if we could trigger to uptest from commonly used resources before moving this PR forward.

Issue #958, opened 2023-11-10
Request for AWS Bedrock resources


blakeromano:
### What resource do you need?

Terraform Resource Name: 

* aws_bedrock_custom_model
* aws_bedrock_provisioned_model_throughput
* aws_bedrock_model_invocation_logging_configuration

### What is your use case?

This is dependent on https://github.com/hashicorp/terraform-provider-aws/issues/33809 

The adoption of AWS Bedrock across various enterprise organizations is growing so Crossplane should be able to provision AWS Bedrock resources.

MichaelRSilva:
/fresh These resources are very important to implement AI at AWS.

PR #1293, opened 2024-04-28
Add AWS Bedrock Provider and `aws_bedrock_model_invocation_logging_configuration` Resource

blakeromano:

This aims to add the `aws_bedrock_model_invocation_logging_configuration` via a new AWS Bedrock provider. This will 
help to address https://github.com/crossplane-contrib/provider-upjet-aws/issues/958 especially once TF provider upgrade is released.

Fixes # https://github.com/crossplane-contrib/provider-upjet-aws/issues/958

I have:


config/bedrock/config.go
```diff
@@ -0,0 +1,24 @@
...
+// Configure adds configurations for the bedrock group.
+func Configure(p *config.Provider) {
+	p.AddResourceConfigurator("aws_bedrock_model_invocation_logging_configuration", func(r *config.Resource) {
+		r.References["logging_config.s3_config.bucket_name"] = config.Reference{
+			Type: "github.com/upbound/provider-aws/apis/s3/v1beta1.Bucket",
+		}
+		r.References["logging_config.cloudwatch_config.role_arn"] = config.Reference{
+			Type:      "github.com/upbound/provider-aws/apis/iam/v1beta1.Role",
+			Extractor: common.PathARNExtractor,
+		}
```
turkenf, line 21: In order for [SingleNestedBlocks](https://github.com/hashicorp/terraform-provider-aws/blob/4ec8ee85f8175b86a217136c120b667a822f481a/internal/service/bedrock/model_invocation_logging_configuration.go#L46) to be generated properly, we need to manually add the following configurations.

Similar example: https://github.com/turkenf/provider-aws/blob/997e72c93954493710b4327098dfbc1928969063/config/opensearchserverless/config.go#L13

```
		r.SchemaElementOptions.SetEmbeddedObject("logging_config")
		r.SchemaElementOptions.SetEmbeddedObject("logging_config.s3_config")
		r.SchemaElementOptions.SetEmbeddedObject("logging_config.cloudwatch_config")
```

config/externalname.go
```diff
@@ -27,6 +27,11 @@ var TerraformPluginFrameworkExternalNameConfigs = map[string]config.ExternalName
 	// Imported by using the id: sgr-02108b27edd666983
 	"aws_vpc_security_group_ingress_rule": vpcSecurityGroupRule(),
 
+	// bedrock
+	//
+	// Bedrock model invocation logging configuration
+	"aws_bedrock_model_invocation_logging_configuration": config.NameAsIdentifier,
```
turkenf, line 33: Since it is imported using ID, it is better to use `config.IdentifierFromProvider` for [external name](https://github.com/crossplane/upjet/blob/v0.10.0/docs/add-new-resource-short.md#external-name-cases).

```suggestion
	// Imported by using the id
	"aws_bedrock_model_invocation_logging_configuration": config.IdentifierFromProvider,
```

examples-generated/bedrock/v1beta1/modelinvocationloggingconfiguration.yaml
```diff
@@ -0,0 +1,70 @@
...
+kind: ModelInvocationLoggingConfiguration
+metadata:
+  annotations:
+    meta.upbound.io/example-id: bedrock/v1beta1/modelinvocationloggingconfiguration
+  labels:
+    testing.upbound.io/example-name: example
+  name: example
+spec:
+  forProvider:
+    region: us-west-1
```
turkenf, line 11: You should add `loggingConfig` block like the following and create the resource example under the `examples/bedrock/v1beta1/modelinvocationloggingconfiguration.yaml`

```suggestion
    region: us-west-1
    loggingConfig:
      embeddingDataDeliveryEnabled: true
      imageDataDeliveryEnabled: true
      s3Config:
        bucketName: example
```

turkenf (commented):
Thank you for your effort in this PR @blakeromano, I left a few comments for you to consider.

turkenf:
Hi @blakeromano,

The `v1.6.0` release is planned for tomorrow, June 6, if you want to include this resource in the release, please look at the comments above, thanks in advance.

config/externalname.go
mbbush: According to the terraform docs, the terraform id is the region (since this resource only supports one per region per account). Using `ParameterAsIdentifier("region")` would remove the region from the schema, which feels like a bad idea as it would violate the existing conventions. 

Given the one per region constraint, I think it's important for this resource to use a "query first based on parameters" type of external name config, allowing an existing configuration to be found and updated even if the external name annotation is not specified. So I think `config.TemplatedStringAsIdentifier("", "{{ .setup.configuration.region }}")` would be my preference. 

@turkenf what do you think?

PR #1310, opened 2024-05-14
Use config.Reference.TerraformName instead of config.Reference.Type

ulucinar:

This PR replaces usages of upjet's `config.Reference.Type` API with `config.Reference.TerraformName`. In https://github.com/crossplane/upjet/pull/400, we are deprecating the `config.Reference.Type` API in favor of `config.Reference.TerraformName`. `config.Reference.TerraformName` is a more stable and less error prone API compared to `config.Reference.Type` because it automatically accounts for the configuration changes affecting the cross-resource reference target's kind name, group or version. We've already been discouraging the `config.Reference.Type` in favor of `config.Reference.TerraformName` since it's been introduced.

This PR is a precursor for converting the singleton list APIs to embedded objects. When the `config.Reference.Type` is used to define the reference target, the API version can be omitted for the targets in the same group and version. In those cases, if the referencing resource has been generated at `v1beta2` whereas the reference target lacks that version, the generated resolver is not valid. Using `config.Reference.TerraformName` instead will help us in future maintenance because it automatically accounts for generating the cross-resource reference resolvers in the generated version. This is also one of the reasons we are deprecating `config.Reference.Type`.

I have:

Checked for any unexpected changes in the generated resolvers (`apis/../zz_generated.resolvers.go` files).

### DIFF
--- a/apis/acm/v1beta1/zz_certificatevalidation_types.go
+++ b/apis/acm/v1beta1/zz_certificatevalidation_types.go
@@ -16,14 +16,14 @@ import (
 type CertificateValidationInitParameters struct {
 
 	// ARN of the certificate that is being validated.
-	// +crossplane:generate:reference:type=Certificate
+	// +crossplane:generate:reference:type=github.com/upbound/provider-aws/apis/acm/v1beta1.Certificate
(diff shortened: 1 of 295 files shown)


sergenyalcin (approved):
Thanks @ulucinar LGTM!

PR #1343, opened 2024-06-05
Remove opsworks provider that is now EOL

### LINKED ISSUE #1224: Remove provider-aws-opsworks once its EOL is complete in May 2024
AWS Opsworks is closed to new customers, and scheduled to become unavailable to all existing customers in the next few months. This means we can remove it from provider-aws, to slightly reduce the size, maintenance, and testing burdens of the provider. 

We should wait until AWS has actually completed their shutdown, as they have extended these types of EOL dates in the past.

Currently, the shutdown schedule is [March 31, 2024 for Puppet](https://docs.aws.amazon.com/opsworks/latest/userguide/opspup-eol-faqs.html), [May 5, 2024 for Chef](https://docs.aws.amazon.com/opsworks/latest/userguide/opscm-eol-faqs.html), and [May 26, 2024 for stacks](https://docs.aws.amazon.com/opsworks/latest/userguide/stacks-eol-faqs.html).

I've never used this service myself, so I don't actually know which managed resources are used for puppet vs chef vs stacks. I'm inclined to wait until all three are shut down, then remove the provider entirely. 

Note that we can also remove the opsworks exception from the `RegionAddition` override. 

Given that the service will no longer exist, although technically its removal would be a breaking change, I'm inclined to release the removal in a normal minor version bump. If there are good reasons not to do so (or reasons to not remove them at all), I'd be happy to hear them.

These resources were originally added in PR #444 (issue #313) . I'm curious if anyone involved in adding them knows if they were specifically requested, or merely added as part of systematically going through all the terraform-provider-aws resources to get broad crossplane support.

mbbush:
On May 26, 2024, AWS completed their deprecation cycle for the opsworks service and it is now completely EOL for both new and existing customers. Nobody can use it anymore. 

This removes the resources from the provider to shrink our maintenance burden, build time, image sizes, etc. 

Fixes #1224

I have:


haarchri:
can we add a message in Release Notes that people can remove the CRDs for opsworks ?

mbbush:
Maybe we release this in the same version that stops shipping the monolithic provider? It's definitely worth mentioning in the release notes, but given what I've learned about opsworks while reading the deprecation docs, I'd be somewhat surprised if _anyone_ is installing `provider-aws-opsworks`.

PR #1349, opened 2024-06-10
Alphabetize external name

mbbush:

`config/externalname.go` is one of the largest and most important hand-written files in this repository. Resources have been added (kind of) wherever in the file. Currently, when looking for a resource, or for several related resources, it is difficult to navigate or know when I've found all of them, and it looks very messy.

Additionally, if I'm adding a new resource, and put it at the bottom, this can create git conflicts with someone else adding an unrelated resource in the same place. 

I'd love to have a linter that automates this, but I couldn't find one that could handle the way we group resources by their go sdk package name, and that keeps the resources together with their explanatory comments. So I just did it manually. 

It is my hope that now that the file is alphabetized, we can enforce that users keep it alphabetized in code review. 

In order to prevent clobbering the git history, I introduced a `.git-blame-ignore-revs` file with the SHA of my rename commit. Once approved, I will merge this with a merge commit to ensure that the sha doesn't change. 

I have:

I programmatically compared the new and old versions of externalname.go and they only differ in comments, ordering of map keys, and whitespace. 

There are new entries in `externalnamenottested.go` because I migrated a few resources that had been commented out in `externalname.go` to uncommented in `externalnamenottested.go`. But we don't currently use that list for anything.

Zero diff to the generated code confirms that this is a pure refactor with no meaningful changes.


turkenf (approved):
Thanks @mbbush, LGTM.

PR #1350, opened 2024-06-10
Make SSO examples uptestable

mbbush:

I noticed when doing https://github.com/crossplane-contrib/provider-upjet-aws/pull/1349 that some of the ssoadmin resources have an external name configuration that uses `index` to extract a value that appears in a singleton object. These seemed like good candidates to test with these changes. 

I'm pleased to report that :heavy_check_mark: the tests pass with no issues for all the v1beta1 and v1beta2 manifests in both identitystore and ssoadmin. Logs are attached. Note that because I deliberately used unique names for each resource when I first wrote these example manifests, I was able to uptest them all at once (per api version), which was quite fast.

I also made the ssoadmin and identitystore resources uptestable by using an uptest datasource.

In order for these to pass with the config used in github actions, someone from upbound will need to:
* Create an instance of aws iam identity center in us-east-1 in the aws organization the uptest account belongs to. If it's not in an organization, then it can just be made into an org (I did this with my personal account)
* If the uptest account is not the primary account for its aws organization, then it needs to be made a delegated admin account in the IAM identity Center console.
* Put the appropriate values in the github secret used for the uptest datasource.

Note on aws terminology: IAM Identiy Center is the current name for what aws originally launched as "sso" and "identity store". We still have the old names in the SDK, but the console uses the new marketing names. 

Fixes #

I have:

- [x] Added `backport release-x.y` labels to auto-backport this PR if necessary.

[ssoadmin-v1beta1.logs.txt](https://github.com/user-attachments/files/15755383/ssoadmin-v1beta1.logs.txt)
[ssoadmin-v1beta2.logs.txt](https://github.com/user-attachments/files/15755384/ssoadmin-v1beta2.logs.txt)
[identitystore.logs.txt](https://github.com/user-attachments/files/15755385/identitystore.logs.txt)

### DIFF
--- a/examples/identitystore/v1beta1/group.yaml
+++ b/examples/identitystore/v1beta1/group.yaml
@@ -7,7 +7,6 @@ kind: Group
 metadata:
   annotations:
     meta.upbound.io/example-id: identitystore/v1beta1/group
-    upjet.upbound.io/manual-intervention: "This resource requires a valid identityStoreId"
   labels:
     testing.upbound.io/example-name: example-group
   name: example-group
@@ -15,5 +14,5 @@ spec:
   forProvider:
     description: Example description
     displayName: Example group
-    identityStoreId: Identity store id from a manually created SSO instance in the same region specified below
+    identityStoreId: "${data.identity_store_id}"
     region: us-east-1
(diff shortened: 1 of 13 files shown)


turkenf:
/test-examples="examples/identitystore/v1beta1/group.yaml"

turkenf:
Thanks for this PR @mbbush,

I added values ​​for `identity_store_id` and `sso_instance_arn` to the data source but I'm not sure if the test account is authorized or not.

PR #1367, opened 2024-06-14
Only add singleton list API converters for the existing APIs

ulucinar:

Only apply the CRD API conversions for converting between singleton list & embedded object APIs for the already existing resources with the generated singleton list CRD APIs.

When we try to add a new Terraform resource with a singleton list in its schema, currently, we need to do a manual configuration such as the following:
```go
p.AddResourceConfigurator("azurerm_pim_active_role_assignment", func(r *config.Resource) {
		r.PreviousVersions = nil
		r.Version = "v1beta1"
		r.Conversions = nil
		r.TerraformConversions = []config.TerraformConversion{
			config.NewTFSingletonConversion(),
		}
	})
```
, so that the API converters registered are removed and versioning starts with `v1beta1` instead of `v1beta2` (for such new resources, the `v1beta1` version does not already exist).

This PR records the names of resources with previous CRD APIs containing (now converted) singleton lists so that the API converters are only registered for them.
This makes adding new resources with singleton lists in their Terraform schemas easier by removing the need to do a manual configuration like shown above.

The newly introduced list should mostly be immutable (it records a set of existing resources with a specific condition at a point in time). Bugs may result in exceptions where we may need to update this list. In other words, new resources should normally be never added to this list, doing so will result in an invalid configuration.

I have:

Via the following uptest runs:
- `NodeGroup.v1beta2.eks` (the `v1beta2` manifest contains a converted embedded object relevant in our context): https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/9518567085
- `DeliveryStream.v1beta2.firehose`: https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/9518491580
- `SecurityConfig.v1beta1.opensearchserverless` (contains an embedded field (`samlOptions`) represented as a single nested block in the Terraform schema): https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/9518549846


ulucinar:
/test-examples="examples/firehose/v1beta2/deliverystream.yaml"

ulucinar:
/test-examples="examples/opensearchserverless/v1beta1/securityconfig.yaml"

ulucinar:
/test-examples="examples/eks/v1beta2/nodegroup.yaml"

config/registry.go
```diff
@@ -34,6 +34,24 @@ var (
...
+	// not possess a CRD API version with the singleton list. Thus, for
+	// the newly added resources (resources added after the singleton lists
+	// have been converted), we do not need the CRD API conversion
+	// functions that convert between singleton lists and embedded objects,
+	// but we need only the Terraform conversion functions.
+	// This list is immutable and represents the set of resources with the
+	// already generated CRD API versions with now converted singleton lists.
+	// Because new resources should never have singleton lists in their
+	// generated APIs, there should be no need to add them to this list.
+	// However, bugs might result in exceptions in the future.
```
mergenci, line 52: Suggestion: Can you add a link to the [PR that introduces singleton list conversion](https://github.com/crossplane-contrib/provider-upjet-aws/pull/1332) here? Doing so would give the reader some context. Otherwise, the reader might be puzzled so as to why such machinery was needed in the first place.

mergenci (approved):
Thanks @ulucinar 🙏 This PR should make new resource configurators' job significantly easier by eliminating the need to understand the shenanigans of singleton list conversions. Please see my suggestion regarding documentation below.

config/registry.go
ulucinar: Done.

ulucinar:
/test-examples="examples/firehose/v1beta2/deliverystream.yaml"

PR #1429, opened 2024-07-30
feat(s3): add s3 directory bucket

haarchri:
implement s3 directory bucket ([aws_s3_directory_bucket](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_directory_bucket))

Fixes #

I have:

- [x] Added `backport release-x.y` labels to auto-backport this PR if necessary.

uptest: https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/10618327577

```
PASS
15:58:46 [ OK ] running automated tests
```


haarchri:
/test-examples="examples/s3/v1beta1/directorybucket.yaml"

config/externalname.go
```diff
@@ -2137,7 +2142,6 @@ var TerraformPluginSDKExternalNameConfigs = map[string]config.ExternalName{
 	"aws_rum_metrics_destination": config.IdentifierFromProvider,
 
 	// s3
-	//
```
turkenf, line 2140: I think there is no need to remove this line for compatibility.

examples/s3/v1beta1/directorybucket.yaml
```diff
@@ -0,0 +1,15 @@
+# https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-express-Endpoints.html
```
turkenf, line 1: Could you please add the license notification?

```suggestion
// SPDX-FileCopyrightText: 2024 The Crossplane Authors <https://crossplane.io>
//
// SPDX-License-Identifier: Apache-2.0

# https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-express-Endpoints.html
```

turkenf (commented):
Thank you @haarchri, left two small comments 🙌

config/externalname.go
```diff
@@ -75,6 +75,11 @@ var TerraformPluginFrameworkExternalNameConfigs = map[string]config.ExternalName
 	// VPCEndpoint can be imported using the AWS-assigned VPC Endpoint ID, i.e. vpce-0a957ae9ed5aee308
 	"aws_opensearchserverless_vpc_endpoint": opensearchserverlessVpcEndpoint(),
 
+	// s3
+	//
+	// S3 directory bucket can be imported using the full id: [bucket_name]--[azid]--x-s3
+	"aws_s3_directory_bucket": config.ParameterAsIdentifier("bucket"),
```
mbbush, line 81: Based on the description in the terraform provider docs, I would expect this to use an external name config that combined the bucket name with the availability zone id from the parameters, and the --x-s3 suffix, to form the terraform id from the external name. I think ideally the external name would be just the bucket name, which you could also make the metadata.name.

examples/s3/v1beta1/directorybucket.yaml
```diff
@@ -0,0 +1,15 @@
...
+apiVersion: s3.aws.upbound.io/v1beta1
+kind: DirectoryBucket
+metadata:
+  annotations:
+    meta.upbound.io/example-id: s3/v1beta1/directorybucket
+    crossplane.io/external-name: ${Rand.RFC1123Subdomain}--usw2-az1--x-s3
```
mbbush, line 7: Expecting users to specify this format on every resource would be a pretty bad user experience, especially since there's no guarantee that the availability zone id in this name would match the location name in the `spec`.

examples/s3/v1beta1/directorybucket.yaml
```diff
@@ -0,0 +1,15 @@
...
+  labels:
+    testing.upbound.io/example-name: example
+  name: example
+spec:
+  forProvider:
+    location:
+    - name: usw2-az1
```
mbbush, line 14: This looks like a singleton list, which I'm surprised to see here. Perhaps the code which converts it into an embedded object only supports SDK resources, not framework?

Especially since this parameter should go into the external name config, I think we should do the necessary work to make this into an embedded object.

config/externalname.go
mbbush: Maybe something like this? The fact that the location name is inside an array in the terraform schema means this may require some tweaking.  
```suggestion
	// S3 directory bucket can be imported using the full id: [bucket_name]--[azid]--x-s3
	"aws_s3_directory_bucket": config.TemplatedStringAsIdentifier("bucket", "{{ .external_name }}--{{ .parameters.location.name }}--x-s3"),
```

haarchri: i couldn't find a possible way because of the array

examples/s3/v1beta1/directorybucket.yaml
mergenci: @mbbush, `location` [is indeed a singleton list](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/s3/directory_bucket.go#L99-L124). Our singleton list conversion implementation processes only Plugin SDK resources. Because Plugin Framework resource schema has a different structure (and type), singleton list detection and conversion should be implemented separately for it.

If I recall correctly though, singleton list fields shouldn't exist in Plugin Framework resources, in the first place, because Plugin Framework schema is able to represent them as embedded objects. Unfortunately, I can't remember why and how at the moment.

mergenci: A clarification… We [traverse schemas of all resources](https://github.com/crossplane/upjet/blob/main/pkg/config/provider.go#L376), Plugin Framework and Plugin SDKv2, but our [singleton list detector](https://github.com/crossplane/upjet/blob/main/pkg/config/schema_conversions.go#L66-L77) doesn't detect singleton lists of Plugin Framework schemas.

config/externalname.go
turkenf: I did not try but the above configuration will not work because, according to the documentation, there is a mandatory format for the `bucket` field.

Screenshot 2024-08-22 at 17 04 55

In this case, we cannot pass a name in the above format without using `crossplane.io/external-name` in `metadata.name`.

turkenf:
> is there any timeline on completing reviews?

Hi @PaulinaZaremba, thank you for your understanding and patience here. We are unable to include PR in this week's release in order to look at the issues mentioned in the reviews in more detail.

We plan to cut a new release next week that includes this PR.

turkenf:
/test-examples="examples/s3/v1beta1/directorybucket.yaml"

examples/s3/v1beta1/directorybucket.yaml
turkenf: We need to make changes on the upjet side to convert the singleton list to embedded object in the framework resources. I've created an issue for [this](https://github.com/crossplane/upjet/issues/430) on upjet, we can follow up from there.

turkenf (approved):
Thanks @haarchri, LGTM.

PR #1456, opened 2024-08-09
Improve external-name config of iot thing type

mbbush:

The aws iot ThingType resource uses a parameter called "name" as its terraform id. Unfortunately, when it was first implemented, it was done using IdentifierFromProvider, which means that creating a new managed resource always attempts to create a new external resource, even if there already is one with the same name. For this particular resource, this results in errors from the AWS api. 

The best external name config would be `config.NameAsIdentifier`, which would remove the redundant `name` property from both `spec.forProvider` and `spec.initProvider`, and initialize the name from `metadata.name`. But that's a breaking api change.

The next best external name config would be `config.TemplatedStringAsIdentifier("", "{{ .parameters.name }}")`, but that will (correctly) set `name` as an identifier field, which will remove it from `spec.initProvider`, which is also a breaking change. 

So this is the best I can do at the moment without changing the schema. The content of the external name remains the same, but now crossplane knows how to construct it from the `spec`, so it can check for resource existence before trying to create it. 

Because `name` is part of the id, but not set as an identifier field, various edge cases where the user omits it may continue to fail in less-than-ideal ways, but that's not a change from current behavior.

Fixes #

I have:

Running uptest on the pr.

Manually confirming that the external name is the in the same format both before and after this change.

### DIFF
--- a/config/externalname.go
+++ b/config/externalname.go
@@ -1512,7 +1512,8 @@ var TerraformPluginSDKExternalNameConfigs = map[string]config.ExternalName{
 	// TODO: For now API is not normalized. While testing resource we can check the actual ID and normalize the API.
 	"aws_iot_thing_principal_attachment": config.IdentifierFromProvider,
 	// IOT Thing Types can be imported using the name
-	"aws_iot_thing_type": config.IdentifierFromProvider,
+	// NameAsIdentifier would be better, but that would be a breaking api change to remove spec.forProvider.name
+	"aws_iot_thing_type": FormattedIdentifierFromProvider("", "name"),
 	// IoT Topic Rules can be imported using the name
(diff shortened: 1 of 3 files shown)


mbbush:
/test-examples="examples/iot/v1beta1/thingtype.yaml"

mbbush:
/test-examples="examples/iot/v1beta2/thingtype.yaml"

mbbush:
/test-examples="examples/iot/v1beta1/thingtype.yaml,examples/iot/v1beta2/thingtype.yaml"

turkenf:
/test-examples="examples/iot/v1beta1/thingtype.yaml"

turkenf (approved):
Thank you @mbbush, LGTM.

PR #1502, opened 2024-09-27
Use different parameters name for auth and config group overrides

sergenyalcin:

This PR introduces a change for parametrizing the auth group and config group overrides. uses different parameters name for auth and config group overrides. There is not a behavior change for the provider because the values are set to the same (PROVIDER_NAME). Now, the dependency between these fields were eliminated.

I have:

Tested locally.


turkenf (approved):
Thank you @sergenyalcin, LGTM.

Issue #881, opened 2023-09-14
upbound aws provider fails to create/delete SNS topic on china due to bad aws partition (shall be aws-cn and not aws)


yebolenko:
### What environment did it happen in?

* Crossplane Version: docker.io/crossplane/crossplane:**v1.13.2**
* Provider Version: xpkg.upbound.io/upbound/provider-aws:**v0.38.0**
* Kubernetes Version:  1.27
* Kubernetes Distribution:  EKS

yebolenko:
Please ensure that all upbound components use aws partition as a parameter and not hardcoded aws

yebolenko:
it seems that some info in the issue has not been update:
I'm summarizing it again:
We create SNS Topic with crossplane/upbound on aws China region cn-north-1
When SNS topic shall be deleted, it fails with the following error:
InvalidParameter: Invalid parameter: TopicArn Reason: A REDACTED ARN must begin with **arn:aws-cn**, not **arn:aws:sns**:REDACTED:REDACTED:ye-sns-topic-2-demo-china-cnn1\\n\\tstatus code: 40I

yebolenko:
any progress?

turkenf:
@yebolenko, could you please fill in the description properly? Reproduction steps, the MRs you used, expected and actual behavior.

yebolenko:
It's always reproducible. You just need to create/delete SNS topic on aws region cn-north-1 (china).
It doesn't work due to hardcoded aws partition id (aws) and not aws-cn. 
I believe it's general issue also in many other resources. You need to add ability to pass aws partition id as parameter (aws is default)

yebolenko:
any update? it's URGENT issue for us. We cannot use SNS and (maybe other components) on china.
As I said, we use an official upbound release

turkenf:
Please provide the requested information in the description.

yebolenko:
Providing again all details:
------------------------- 
**What happened?**  When SNS topic shall be created/deleted, it fails with the following error:
**InvalidParameter: Invalid parameter: TopicArn Reason: A REDACTED ARN must begin with arn:aws-cn, not arn:aws:sns:REDACTED:REDACTED:ye-sns-topic-2-demo-china-cnn1\n\tstatus code: 40I**
How can we reproduce it?  It's always reproducible. You just need to create/delete SNS topic on **aws region cn-north-1 (china).**
What environment did it happen in?
[...]

Issue #820, opened 2023-08-09
Provider-AWS does not support GovCloud Partition IDs


rsavage-nozominetworks:
### What happened?

I have multiple AWS accounts, some in the regular `Commercial` regions, and some in the `Gov Cloud` regions.  All of my composite resources create as expected in the `Commercial` regions.   As for the ones in the `Gov Cloud` regions, I am seeing issues where certain AWS resources aren't able to create due to improper partition ids.   In this example, I am attempting to create a SNS topic which will alert for relative issues.  I am providing my provider and controller configs below.

### How can we reproduce it?

Original provider configuration (before any attempts to troubleshoot):
```yaml
apiVersion: aws.upbound.io/v1beta1
kind: ProviderConfig
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"aws.upbound.io/v1beta1","kind":"ProviderConfig","metadata":{"annotations":{},"name":"default"},"spec":{"credentials":{"source":"IRSA"}}}
  finalizers:
  - in-use.crossplane.io
  generation: 1
  name: default
spec:
  credentials:
    source: IRSA
status:
  users: 9
  ```
  An updated provider configuration (after trying to troubleshoot):
```yaml
apiVersion: aws.upbound.io/v1beta1
kind: ProviderConfig
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"aws.upbound.io/v1beta1","kind":"ProviderConfig","metadata":{"annotations":{},"name":"default"},"spec":{"credentials":{"source":"IRSA"}}}
  finalizers:
  - in-use.crossplane.io
  generation: 2
  name: default
spec:
  credentials:
    source: IRSA
  endpoint:
    partitionId: aws-us-gov
    url:
      dynamic:
        host: amazonaws-us-gov.com
        protocol: https
      type: Dynamic
status:
  users: 9
```

Here is my current ControllerConfig.   Notice: I added the env variables for AWS_REGION and AWS_DEFAULT_REGION as recommend by [other related issues](https://github.com/upbound/provider-aws/issues/757).
```yaml
apiVersion: pkg.crossplane.io/v1alpha1
kind: ControllerConfig
metadata:
  annotations:
    eks.amazonaws.com/role-arn: arn:aws-us-gov:iam::REDACTED:role/REDACTED-crossplane
[...]

PR #1223, opened 2024-03-18
Use the configured AWS partition when constructing ARNs

mbbush:
### Description of your changes

Stores the existing partition id from the provider config into a terraform configuration object, which is available when resolving external names. 

Uses the partition as part of the external name config for the resources that currently hardcode a partition of `aws` in their templated ARN.

Refactors the external name configurations which create arns to use simple helper functions, to reduce code repetition.

The default value of the partition is "aws", which is currently the only supported value, so this should be a nonbreaking change for users. 

Fixes #881 
Fixes #820 
Addresses #855 (although probably not completely)

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable` to ensure this PR is ready for review.
- [ ] Added `backport release-x.y` labels to auto-backport this PR if necessary.

### How has this code been tested

[contribution process]: https://git.io/fj2m9

Issue #855, opened 2023-08-31
Improve support for AWS GovCloud


jeanduplessis:
### What problem are you facing?

Support for AWS GovCloud is lacking in the provider today. We need to implement proper support for it.

Examples of issues users have faced so far:
- https://github.com/upbound/provider-aws/issues/820
- https://github.com/upbound/provider-aws/issues/757

### How could Official AWS Provider help solve your problem?

Implement proper support for GovCloud.

Consider looking at how `crossplane-contrib/provider-aws` deals with this:
- https://github.com/crossplane-contrib/provider-aws/pull/1329
- https://github.com/crossplane-contrib/provider-aws/pull/1364

mbbush:
One difficulty in supporting this using upjet is that when writing an ExternalName configuration, the region is available, but not the partition. 

What if we had an `ARNBuilder` function or something, for use in constructing ARNs, which knew which regions are in which partitions? That could also help DRY up some of the string concatenation used in external name configurations, where we're currently assuming the aws-commercial partition, for e.g. [iam policy](https://github.com/upbound/provider-aws/blob/c1f16051f733c89cac2d7b5c60d337370a1f3ce1/config/externalname.go#L2686), [sns topic](https://github.com/upbound/provider-aws/blob/c1f16051f733c89cac2d7b5c60d337370a1f3ce1/config/externalname.go#L863), [elasticsearch domain](https://github.com/upbound/provider-aws/blob/c1f16051f733c89cac2d7b5c60d337370a1f3ce1/config/externalname.go#L2388C5-L2388C5),   [state functions](https://github.com/upbound/provider-aws/blob/c1f16051f733c89cac2d7b5c60d337370a1f3ce1/config/externalname.go#L1194-L1195), and others

jacob-kuder:
Any update on this?

PR #1554, opened 2024-11-04
[revive] Use the configured AWS partition when constructing ARNs

erhancagirici:

Revives the the original PR from @mbbush #1223
> - Stores the existing partition id from the provider config into a terraform configuration object, which is available when resolving external names.
> - Uses the partition as part of the external name config for the resources that currently hardcode a partition of aws in their templated ARN.
> - Refactors the external name configurations which create arns to use simple helper functions, to reduce code repetition.
> - The default value of the partition is "aws", which is currently the only supported value, so this should be a nonbreaking change for users.
> Fixes https://github.com/crossplane-contrib/provider-upjet-aws/issues/881
> Fixes https://github.com/crossplane-contrib/provider-upjet-aws/issues/820
> Addresses https://github.com/crossplane-contrib/provider-upjet-aws/issues/855 (although probably not completely)
On top of the original PR:
- rebased onto latest main
- added aws-iso-e partition

- default regions for IAM MRs are now generated from AWS service definitions, instead of being hardcoded. 
  - `make generate` will be handling the generation.
- Introduced a new accepted value `Auto` for `spec.endpoint.url.type` in `ProviderConfig`. This allows specifying a partition in `spec.endpoint.partitionId` without having to specify a full endpoint configuration. Relevant AWS endpoints will be automatically resolved.
- MRs with `region` parameter will use the `region` to infer the `partition` without extra config in the `ProviderConfig`.
  - if a partition is configured in the `ProviderConfig` and the `region` provided by the resource conflicts, this conflict will intentionally cause an error during reconcile.
- For Region-less MRs (like MRs in `iam.aws.upbound.io`) to work with non-default partitions, referenced `ProviderConfig` must specify the `partitionId`, an example for `aws-cn` partition is:
```yaml
kind: ProviderConfig
metadata:
  name: default
spec:
  credentials:
    secretRef:
      key: creds
      name: aws-secret
      namespace: upbound-system
    source: Secret
  endpoint:
    partitionId: aws-cn
    url:
      type: Auto
```
This `ProviderConfig` with explicit partition config is the recommended way to work with non-default AWS partitions.

> Note: Consumers that operate on default AWS partition do not need to take any action.

I have:

- via uptest: regression tests for the default AWS partition
- e2e tests with a custom image out of this PR: contributors with `aws-cn` and `aws-us-gov` access. Special thanks to them!

---
Resource Name | Test results for the default AWS partition
-- | --
route.ec2.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12949524630
eip.ec2.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12995158533
internetgateway.ec2.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12955462687
mainroutetableassociation.ec2.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12949548200
natgateway.ec2.aws.upbound.io | manual: https://github.com/crossplane-contrib/provider-upjet-aws/pull/1554#issuecomment-2616573170
routetableassociation.ec2.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12949559061
routetable.ec2.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12949559061
securitygroupingressrule.ec2.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12950670145
securitygrouprule.ec2.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12950679267
securitygroup.ec2.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12950688596
subnet.ec2.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12955461354
vpc.ec2.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12949504090
addon.eks.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12950711544
clusterauth.eks.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12953858109
cluster.eks.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12953858109
openidconnectprovider.iam.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12955459497
policy.iam.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12950741168
rolepolicyattachment.iam.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12953867470
role.iam.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12953867470
instance.rds.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12950755421
subnetgroup.rds.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12950763216
bucketlifecycleconfiguration.s3.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12950771123
bucketlogging.s3.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12950783823
bucketpolicy.s3.aws.upbound.io | manual: https://github.com/crossplane-contrib/provider-upjet-aws/pull/1554#issuecomment-2616573170
bucket.s3.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12950805510
bucketversioning.s3.aws.upbound.io | https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/12950796393


mbbush (approved):
LGTM. Thanks for picking this up!

internal/clients/aws.go
```diff
@@ -38,6 +39,16 @@ type SetupConfig struct {
 	Logger            logging.Logger
 }
 
+// iamRegions holds the region used for signing IAM credentials for each AWS partition.
+var iamRegions = map[string]string{
```
ulucinar, line 43: What's the source of truth for this table?

Following is a more robust implementation for the signing region lookup keyed by the partition ID:
```go
	// map from a partition to its IAM signing region
	signingRegionMap := make(map[string]string)
	for _, p := range endpoints.DefaultResolver().(endpoints.EnumPartitions).Partitions() {
		for _, r := range p.Regions() {
			re, err := p.EndpointFor("iam", r.ID(), func(opt *endpoints.Options) {
				opt.ResolveUnknownService = true
			})
			if err != nil {
				panic(err)
			}
			signingRegionMap[p.ID()] = re.SigningRegion
			break
		}
	}
```

It currently produces the following map:
```
map[aws:us-east-1 aws-cn:cn-north-1 aws-iso:us-iso-east-1 aws-iso-b:us-isob-east-1 aws-iso-e:eu-isoe-west-1 aws-us-gov:us-gov-west-1]
```

Please note the difference for the partition `aws-cn`, don't know if this really matters.

Please also see the discussion below regarding partition lookups using the `region` as a key, which are also applicable here.

Also please note that we have `spec.endpoint.signingRegion` in the `ProviderConfig` API, which we should factor into this implementation. But I suggest we do it with a follow-up PR.

internal/clients/aws.go
```diff
@@ -127,11 +144,22 @@ func getAWSConfigWithDefaultRegion(ctx context.Context, c client.Client, obj run
...
-		cfg.Region = "us-east-1"
+		cfg.Region = getIAMRegion(pc)
 	}
 	return cfg, nil
 }
 
+func getIAMRegion(pc *v1beta1.ProviderConfig) string {
+	defaultRegion := "us-east-1"
+	if pc == nil || pc.Spec.Endpoint == nil || pc.Spec.Endpoint.PartitionID == nil {
+		return defaultRegion
```
ulucinar, line 155: Please see the comment regarding the `iamRegions` map above. We should not be using `us-east-1` for the partitions other than `aws` and this implementation leaves the burden of configuring the right partition for an IAM resource to the user by specifying a proper `ProviderConfig`. While we should allow for overrides via the `ProviderConfig` (as it's already part of the `ProviderConfig` API, and for cases where the SDK has not updated its known partitions), we can provide a better UX with a robust & low maintenance cost implementation that relies on the `aws-go-sdk`.

config/externalname.go
ulucinar, line 1: I feel like the most risky changes in this PR are the conversions of the hardcoded ARNs in the external names to the calls to `fullArnTemplate`. But the conversion is rather straightforward and we can trigger uptest on the resources with external-name implementation changes, where applicable.

Not probably for this PR, but we could also have a unit test that asserts the expected external-name syntax for the configured resources by invoking the implemented external-name configuration (and probably other aspects of the configuration). cc. @turkenf

config/externalname.go
```diff
@@ -3228,3 +3228,36 @@ func eksPodIdentityAssociation() config.ExternalName {
 	}
 	return e
 }
+
+// fullArnTemplate builds a templated string for constructing a terraform id component which is an ARN, which includes
+// the aws partition, service, region, account id, and resource. This is by far the most common form of ARN.
+// e.g. arn:aws:ec2:ap-south-1:123456789012:instance/i-1234567890ab
+func fullArnTemplate(service string, resource string) string {
```
ulucinar, line 3235:
```suggestion
func fullARNTemplate(service string, resource string) string {
```

config/externalname.go
```diff
@@ -3228,3 +3228,36 @@ func eksPodIdentityAssociation() config.ExternalName {
...
+// e.g. arn:aws:iam::123456789012:role/example
+func regionlessArnTemplate(service string, resource string) string {
+	return genericArnTemplate(service, resource, false, true)
+}
+
+// genericArnTemplate builds a templated string for constructing a terraform id component which is an ARN of any format.
+// It always includes the aws partition, service, and resource. Unless you specify to elide them, it will also include
+// templates which resolve to the region (from the spec.forProvider) and the account id (calculated from the provider
+// config).
+func genericArnTemplate(service string, resource string, elideAccountId bool, elideRegion bool) string {
```
ulucinar, line 3252:
```suggestion
func genericARNTemplate(service string, resource string, elideAccountId bool, elideRegion bool) string {
```

ulucinar, line 3252: nit: Looks like in all calls to `genericArnTemplate`, the `elideAccountId` is set to `false`. We can eliminate the parameter till we actually need it.

config/externalname.go
```diff
@@ -3228,3 +3228,36 @@ func eksPodIdentityAssociation() config.ExternalName {
...
+	region := "{{ .setup.configuration.region }}"
+	if elideRegion {
+		region = ""
+	}
+	accountId := "{{ .setup.client_metadata.account_id }}"
+	if elideAccountId {
+		accountId = ""
+	}
+	partition := "{{ .setup.configuration.partition }}"
+	return fmt.Sprintf("arn:%s:%s:%s:%s:%s", partition, service, region, accountId, resource)
```
ulucinar, line 3262: nit:
```suggestion
	return fmt.Sprintf("arn:{{ .setup.configuration.partition }}:%s:%s:%s:%s", service, region, accountId, resource)
```

ulucinar (commented):
Hi @erhancagirici,
Left some comments to provide a better UX while keeping the implementation more robust and easy to maintain, and some other nits.

ulucinar:
Hmm, it's not straightforward to port this code to `aws-sdk-go-v2`, as the `endpoints` package has been removed...

internal/clients/aws.go
```diff
@@ -93,7 +104,13 @@ func SelectTerraformSetup(config *SetupConfig) terraform.SetupFn { // nolint:goc
 		}
 		ps.ClientMetadata = map[string]string{
 			keyAccountID: credCache.accountID,
+			keyPartition: "aws",
 		}
+
+		if pc.Spec.Endpoint != nil && pc.Spec.Endpoint.PartitionID != nil {
+			ps.ClientMetadata[keyPartition] = *pc.Spec.Endpoint.PartitionID
+		}
```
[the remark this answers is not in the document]
erhancagirici: this was also in my mind. I was a bit hesitant to source partitions from `github.com/aws/aws-sdk-go` as it has a deprecation notice [here](https://github.com/aws/aws-sdk-go#warning-this-sdk-is-in-maintenance-mode), stating

> Per that announcement, as of 7/31/2024, the SDK has entered maintenance mode. Going forward, we will limit releases to address critical bug fixes and security issues only. The SDK will not receive API updates for new or existing services, or be updated to support new regions.
Maintenance mode will last for 1 year. After 7/31/2025, the SDK will enter end-of-support, and no updates at all will be made. We recommend that you migrate to [AWS SDK for Go v2](https://aws.github.io/aws-sdk-go-v2/docs/). For additional details as well as information on how to migrate, please refer to the linked announcement.

Searched for an equivalent in `aws-sdk-go-v2` but relevant parts were moved to internal packages, (if I did not miss anything) I could not find anything exported for this.

probably for short term this won't be an issue, but I thought that we can implement something for now to enable the feature, and iterate for cheaper maintenance code.

erhancagirici:

erhancagirici:

erhancagirici:
/test-examples="examples/sns/v1beta1/topic.yaml"

erhancagirici:
/test-examples="examples/iam/v1beta1/policy.yaml"

ulucinar:
Btw, I saw there's also an [ec2.DescribeRegions](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeRegions.html) operation. I gave it a try with the `AllRegions` option but it does not know about other partitions.

ulucinar:
As a pointer to the codegen approach, here's a reference:
```shell
> curl -sL "https://raw.githubusercontent.com/aws/aws-sdk-go-v2/061540b5a73ead172928c9c5aebc48c011bf4ee1/codegen/smithy-aws-go-codegen/src/main/resources/software/amazon/smithy/aws/go/codegen/endpoints.json" | jq '.partitions[].partition'

"aws"
"aws-cn"
"aws-us-gov"
"aws-iso"
"aws-iso-b"
"aws-iso-e"
"aws-iso-f"
```

erhancagirici:
> Looks like the source of truth for the region -> partition mapping can be this file: https://github.com/aws/aws-sdk-go-v2/blob/061540b5a73ead172928c9c5aebc48c011bf4ee1/codegen/smithy-aws-go-codegen/src/main/resources/software/amazon/smithy/aws/go/codegen/endpoints.json

Yes, discovered terraform's `aws-sdk-base` utilizes some code generation based on that JSON. 
[generator](https://github.com/hashicorp/aws-sdk-go-base/blob/82665eea5027433455fc960dac2a7e1140e212b4/internal/generate/endpoints/main.go#L70)
[template](https://github.com/hashicorp/aws-sdk-go-base/blob/82665eea5027433455fc960dac2a7e1140e212b4/internal/generate/endpoints/output.go.gtpl#L13)
[generated](https://github.com/hashicorp/aws-sdk-go-base/blob/82665eea5027433455fc960dac2a7e1140e212b4/endpoints/endpoints_gen.go#L12)

`aws-sdk-go-base` also exposes various helper funcs such as `PartitionForRegion` https://github.com/hashicorp/aws-sdk-go-base/blob/82665eea5027433455fc960dac2a7e1140e212b4/endpoints/partition.go#L64.

we can follow some generation approach too or directly consume `aws-sdk-go-base`

internal/clients/aws.go
erhancagirici: @ulucinar while I understand the comment about `iamRegions` map, I don't think it is directly relevant for the particular case of IAM resources. Note that this code part is only called for IAM MRs.

> this implementation leaves the burden of configuring the right partition for an IAM resource to the user by specifying a proper ProviderConfig

IIUC, unlike other resources, we anyways need some "partition-like" input at the `ProviderConfig` level as IAM resources are "regionless", so there is nothing else from which we can "infer" the partition, by only looking at the MR. Configuring a partition is actually a requirement rather than a config burden.

Alternative can be introducing an optional "partition" input for IAM MR specs, but I don't think this has a big benefit. Because the associated credentials of the provider config are "partition-bound" as described [here](https://docs.aws.amazon.com/whitepapers/latest/aws-fault-isolation-boundaries/partitions.html), i.e. they will be only valid for a single particular partition. The credentials & ProviderConfig actually "dictates" the partition already. 

Unlike handling of regions, where it is possible to use the same provider config and switch regions at MR, this does not make sense for partitions.

internal/clients/aws.go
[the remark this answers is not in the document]
mbbush: What Erhan described is pretty much exactly my original thinking when I first wrote this. It can be done with the go sdk v1, but not v2, and given the pending deprecation I didn't want to add a new dependency on v1.

zonybob:
Curious if this pull request still has hope. We currently can't get the provider to work in our non-standard region due to this issue. Is there an issue with the proposed approach? Or is it just a lack of time and contributors? If it's the latter, I may be able to invest some time. Thanks!

jeanduplessis:
@zonybob its mostly are validation. We don't have access to these non-standard regions which makes validation difficult.
If you're willing to help validate a test image we would very much appreciate it.

zonybob:
> @zonybob its mostly around validation. We don't have access to these non-standard regions which makes validation difficult. If you're willing to help validate a test image we would very much appreciate it.

Yes, I'd be willing to help you validate! Depending on which region I can access to test, it could take me a few days to get to.

mergenci:
Following two issues seem to me that they would be fixed by this PR, though I haven't examined them in detail:
- https://github.com/crossplane-contrib/provider-upjet-aws/issues/1575
- https://github.com/crossplane-contrib/provider-upjet-aws/issues/820

zonybob:
I can test the current pull request on GovCloud. What's the most advisable way to get a built image?

turkenf:
> I can test the current pull request on GovCloud. What's the most advisable way to get a built image?

@zonybob, please see: https://github.com/crossplane/crossplane/discussions/3817#discussioncomment-6428165

internal/clients/aws.go
erhancagirici: As suggested, default IAM region map is now being generated from the source of truth in aws-sdk-go-v2 [here](https://raw.githubusercontent.com/aws/aws-sdk-go-v2/main/codegen/smithy-aws-go-codegen/src/main/resources/software/amazon/smithy/aws/go/codegen/endpoints.json)

internal/clients/aws.go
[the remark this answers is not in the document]
erhancagirici: as suggested, now we try to determine the partition from the supplied region from MR. 
- If we have an explicit `spec.endpoint.partitionId` specified in the `ProviderConfig`, we fall back to that.
- if the MR specified region and `spec.endpoint.partitionId` conflicts, we error out.
- regionless MRs (iam group) relies on `partitionId` from the `ProviderConfig` regardless of the situation, as there is no other way to infer their intended partition.

erhancagirici:
> I can test the current pull request on GovCloud. What's the most advisable way to get a built image?

@zonybob if you want to build your own images be sure to generate the family provider itself along with needed service providers.
e.g. for `iam` `ec2` and `config` (the family provider)
```
make SUBPACKAGES="config iam ec2" BRANCH_NAME=main VERSION=v1.20.0-dev.fe08da0ac-partitions XPKG_REG_ORGS=index.docker.io/erhancag build.all publish
```

I have pre-built images out of this PR, for `iam` and `ec2` if you want to use them:
```
index.docker.io/erhancag/provider-family-aws:v1.20.0-dev.fe08da0ac-partitions
index.docker.io/erhancag/provider-aws-iam:v1.20.0-dev.fe08da0ac-partitions
index.docker.io/erhancag/provider-aws-ec2:v1.20.0-dev.fe08da0ac-partitions
```

Your provider packages should look something similar to this
Family provider
```yaml
apiVersion: pkg.crossplane.io/v1
kind: Provider
metadata:
  name: provider-family-aws
spec:
  package: index.docker.io/erhancag/provider-family-aws:v1.20.0-dev.fe08da0ac-partitions
  skipDependencyResolution: true
```
IAM provider
```yaml
apiVersion: pkg.crossplane.io/v1
kind: Provider
metadata:
  name: provider-aws-iam
spec:
  package: index.docker.io/erhancag/provider-aws-iam:v1.20.0-dev.fe08da0ac-partitions
  skipDependencyResolution: true
```
your provider config should look sth like:
```yaml
kind: ProviderConfig
metadata:
  name: default
spec:
  credentials:
     ...YOUR CREDENTIAL CONFIGURATION...
  endpoint:
    partitionId: aws-us-gov
    url:
      type: Auto
```

turkenf:

turkenf:
/test-examples="examples/ec2/v1beta1/internetgateway.yaml"

turkenf:
/test-examples="examples/ec2/v1beta1/mainroutetableassociation.yaml"

turkenf:
/test-examples="examples/ec2/v1beta1/routetableassociation.yaml"

turkenf:
/test-examples="examples/ec2/v1beta1/securitygroupegressrule.yaml"

turkenf:
/test-examples="examples/ec2/v1beta1/securitygrouprule.yaml"

turkenf:
/test-examples="examples/ec2/v1beta1/securitygroup.yaml"

turkenf:
/test-examples="examples/ec2/v1beta1/subnet.yaml"

turkenf:
/test-examples="examples/eks/v1beta1/addon.yaml"

turkenf:
/test-examples="examples/iam/v1beta1/openidconnectprovider.yaml"

turkenf:
/test-examples="examples/iam/v1beta1/policy.yaml"

turkenf:
/test-examples="examples/rds/v1beta3/instance.yaml"

turkenf:
/test-examples="examples/rds/v1beta1/subnetgroup.yaml"

turkenf:
/test-examples="examples/s3/v1beta2/bucketlifecycleconfiguration.yaml"

turkenf:
/test-examples="examples/s3/v1beta2/bucketlogging.yaml"

turkenf:
/test-examples="examples/s3/v1beta2/bucketversioning.yaml"

turkenf:
/test-examples="examples/s3/v1beta2/bucket.yaml"

turkenf:
/test-examples="examples/iam/v1beta1/openidconnectprovider.yaml"

turkenf:
/test-examples="examples/ec2/v1beta1/subnet.yaml"

turkenf:
/test-examples="examples/ec2/v1beta1/internetgateway.yaml"

turkenf:
/test-examples="examples/eks/v1beta1/clusterauth.yaml"

turkenf:

turkenf:
/test-examples="examples/iam/v1beta1/openidconnectprovider.yaml"

turkenf:
/test-examples="examples/ec2/v1beta1/subnet.yaml"

turkenf:
/test-examples="examples/ec2/v1beta1/internetgateway.yaml"

turkenf:
/test-examples="examples/s3/v1beta2/bucket.yaml"

turkenf:
/test-examples="examples/batch/v1beta1/schedulingpolicy.yaml"

turkenf:
/test-examples="examples/ecs/v1beta2/cluster.yaml"

turkenf:
/test-examples="examples/elasticsearch/v1beta1/domain.yaml"

turkenf:
/test-examples="examples/networkmanager/v1beta1/customergatewayassociation.yaml"

zonybob:
I have run a couple of examples in aws-us-gov, and I can see them all become Ready/Synced=True. They also have proper ARNs. Is there a way you'd like me to help report test results?

turkenf:
> I have run a couple of examples in aws-us-gov, and I can see them all become Ready/Synced=True. They also have proper ARNs. Is there a way you'd like me to help report test results?

Hi @zonybob, 
Thank you for testing and sharing the results 🙏  Your confirmation is sufficient for validation at this point, so no need to report further details. Once we complete the tests in the default region, we'll proceed with the PR. Your help is greatly appreciated! 🙌

turkenf:
/test-examples="examples/elasticsearch/v1beta1/domain.yaml"

sergenyalcin (approved):
@erhancagirici @turkenf @mbbush thank you all very much for the great effort here, it is a very impressive development for the provider-upjet-aws. LGTM!

PR #1602, opened 2024-12-12
Read external ID for assume role authentication from k8s Secret

### LINKED ISSUE #1548: ProviderConfig externalID should take a secretRef

We use an external-id to assume through our auth account into other accounts. This value is, and should be treated as, a secret. Currently in the `ProviderConfig` you can only hard code a value into the resource, which is a problem when they're in gitops.

### What could help solve your problem?

Change structure from
```
spec:
  assumeRoleChain:
  - externalID: <string>
```
to
```
spec:
  assumeRoleChain:
  - externalID:
      secretRef:            # Either this
        name: <string>
        namespace: <string>
        key: <string>
      value: <string>       # or this
```

baburciu:

Fixes #1548

I have:

Adds the `ProviderConfig` ability to define the external ID presented by third parties to access to AWS accounts via:
```yaml
spec:
  assumeRoleChain:
  - externalIDSecretRef:
      namespace: ..
      name: ..
      key: ..
```
on top of existing functionality
```yaml
spec:
  assumeRoleChain:
  - externalID: ..
```
and if both are used, the value defined through `externalIDSecretRef` takes precedence (considered more secure).

A function to read from a Kubernetes Secret was needed and seeing [`ExtractSecret ` in `crossplane-runtime/apis/common/v1`](https://github.com/crossplane/crossplane-runtime/blob/19d95a69cc03690c4b867ff91d89681fcf872a93/pkg/resource/providerconfig.go#L77-L87) I thought I could piggyback on that, to avoid defining another in `provider-upjet-aws`. Maybe a better way would be to allow something more generic in `crossplane-runtime` (today it expects a [`CommonCredentialSelectors`](https://github.com/crossplane/crossplane-runtime/blob/19d95a69cc03690c4b867ff91d89681fcf872a93/apis/common/v1/resource.go#L260-L277))
I don't have much Go development experience, please feel free to suggest if a better approach could be used. 

Trying to run `make reviewable` takes a lot of time, can't say I passed that step.

Tested in an EKS cluster where the provider is using [IRSA](https://docs.upbound.io/providers/provider-aws/authentication/#iam-roles-for-service-accounts) for authentication
```shell
$ kubectl get userpoolclient.cognitoidp.aws.upbound.io demo-mr-externalidsecretref
NAME                                        SYNCED   READY   EXTERNAL-NAME                AGE
demo-mr-externalidsecretref   True     True    7g4kcls3pv3373noulem3gs378   3h5m
$ kubectl get userpoolclient.cognitoidp.aws.upbound.io demo-mr-externalidsecretref -o yaml | yq .spec.providerConfigRef.name
provider-aws-upbound-cognitoidp
$ kubectl get providerconfig.aws.upbound.io provider-aws-upbound-cognitoidp -o yaml | yq .spec
assumeRoleChain:
- externalIDSecretRef:
    key: externalid
    name: externalid-secret
    namespace: crossplane-system
  roleARN: arn:aws:iam::112233445566:role/eks-crossplane-cognito
credentials:
  source: IRSA
$
```
with
```shell
$ kubectl create secret generic externalid-secret -n crossplane-system --from-literal externalid=foobartoto
```
and the assumed role having a trust relationship like:
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": [
                    "arn:aws:iam::778899101011:root"
                ]
            },
            "Action": "sts:AssumeRole",
            "Condition": {
                "StringEquals": {
                    "sts:ExternalId": "foobartoto"
                }
            }
        }
    ]
}
```
(and provider identity allowed to `sts:AssumeRole`).

Similarly tested in kind with [provider authentication method Secret](https://docs.upbound.io/providers/provider-aws/authentication/#create-a-providerconfig), all testing done using dirty builds like
```shell
xpkg pushed to index.docker.io/bogdanadrianburciu/provider-aws-cognitoidp:v1.18.0-rc.0.24.g9a1e8e066.dirty
xpkg pushed to index.docker.io/bogdanadrianburciu/provider-family-aws:v1.18.0-rc.0.24.g9a1e8e066.dirty
```


internal/clients/provider_config.go
```diff
@@ -450,10 +450,38 @@ func UseUpbound(ctx context.Context, region string, pcs *v1beta1.ProviderConfigS
...
+		if aro.ExternalID != nil {
+			opt.ExternalID = aro.ExternalID
+		}
+		if aro.ExternalIDSecretRef != nil {
+			// Fetch secret value
+			secretValue, err := ExtractSecretValue(ctx, kube, aro.ExternalIDSecretRef)
+			if err != nil {
+				fmt.Printf("Error fetching ExternalID from secret: %v\n", err)
+				return
+			}
```
mbbush, line 487: I'm not sure if this is the right way to handle this error here. As I read it, if there's an error fetching the external id from the secret, then the resulting provider config will be missing its tags config. That doesn't sound right.

mbbush:
closing and reopening to re-trigger CI

mbbush (commented):
This looks like a reasonable, non-breaking change. I'm sorry it sat unreviewed for so long. 

I don't think the error handling is quite right, and it looks like there are some probably-unintentional changes in the build submodule that you should clean up. A clean rebase off of the latest main would also be appreciated (I did a merge from the github GUI just to get CI to run). But these are all minor issues.

internal/clients/provider_config.go
baburciu: hi @mbbush, thank you for the review and pointers. I have rebased the PR and addressed the `return` premature exit and  build submodule sync points in separate commits. I can squash those once you'll have the chance to have a look, if it's ok.

internal/clients/provider_config.go
```diff
@@ -457,10 +457,38 @@ func UseUpbound(ctx context.Context, region string, pcs *v1beta1.ProviderConfigS
...
+		}
+		if aro.ExternalIDSecretRef != nil {
+			// Fetch secret value
+			secretValue, err := ExtractSecretValue(ctx, kube, aro.ExternalIDSecretRef)
+			if err != nil {
+				fmt.Printf("Error fetching ExternalID from secret: %v\n", err)
+			} else {
+				secretValueStr := string(secretValue)
+				opt.ExternalID = &secretValueStr
+			}
```
mbbush, line 489: This looks good. As I read it, the only behavior change (besides the desired one) would be that if `opt.ExternalID` is set to a non-nil value, `aro.ExternalId` is nil, and there's an error retrieving the secret, the provider will keep whatever the existing value is instead of setting it to nil. That seems reasonable. 

I'm not super familiar with general golang conventions, so I ask this without any bias one way or the other: Is it stylistically preferred to write
```
if err == nil {
...
} else {
...
}
```
or to use `err != nil`?

I don't personally feel strongly either way.

baburciu: hi @mbbush, I don't have much Go coding experience so not really sure either. 
I do know that early return is always preferred, but for this method we need to only override `opt.ExternalID` with what we read from a k8s Secret only when there's no error, hence we don't get to return, so moving
```go
	if err != nil {
		// log that we couldn't read from k8s Secret and reason, implicitly keeping value aro.ExternalID 
	} else {
		// override value with Secret contents
	}
```
to
```go
	if err == nil {
               // override value with Secret contents
	} else {
               // log that we couldn't read from k8s Secret and reason, implicitly keeping value aro.ExternalID 		
	}
```
could emphasize the "happy path" first and make it more clear for the reader. 
I'll go for your intuition and change to that, so that we leave the `if err != nil` occurrences in the code for their usual early return

baburciu: changed in the last commit and rebased, as it still looked like I'm missing something in the [check-diff](https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/13772950855/job/39410353685?pr=1602#logs) job. If it's something you could easily fix with a `make generate`, please feel free to push on top of this, or please do tell if I need to have another try at it

PR #1721, opened 2025-03-05
Update DynamoDB GSI rangeKey description

apnsngr:

Removed "must be defined" from the DynamoDB global secondary index range key description. This text makes it sound like range key is a required field when it is not. This update makes the description match the local secondary index.

I have:

I have spent several hours trying to run `make reviewable`. It can't get it to work with a clean checkout of the main branch. I'm willing to run this, but I will need some instructions to help me establish a working development environment. I have tried with go 1.23.6, installing goimports, and trying many of the make targets.

This is just a documentation change.

### DIFF
--- a/package/crds/dynamodb.aws.upbound.io_tables.yaml
+++ b/package/crds/dynamodb.aws.upbound.io_tables.yaml
@@ -145,7 +145,7 @@ spec:
                             to the attributes that thatKEYS_ONLY project.
                           type: string
                         rangeKey:
-                          description: Name of the range key; must be defined
+                          description: Name of the range key.
                           type: string
                         readCapacity:
                           description: Number of read units for this index. Must be
@@ -492,7 +492,7 @@ spec:
                             to the attributes that thatKEYS_ONLY project.
                           type: string
                         rangeKey:
-                          description: Name of the range key; must be defined
+                          description: Name of the range key.
                           type: string
                         readCapacity:
                           description: Number of read units for this index. Must be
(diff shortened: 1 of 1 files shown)


mbbush:
Hi @apnsngr, thanks for the PR!

Unfortunately, the files you've updated are all generated code, which is why the `check-diff` check is failing. 

[Here's one example of how to adjust the description of a field](https://github.com/crossplane-contrib/provider-upjet-aws/blob/45ec02187ac6371229b719fc3eb277914a666552/config/kafka/config.go#L234). Once you make a similar change to the file in `config/dynamodb/config.go`, run `make generate` to see the results. It's normal for `make generate` to take 10-15 minutes to run, much of that time without producing output, so it's a good time for a coffee break 😉. 

I hear you on the difficulty of running `make reviewable` locally. It's something I've struggled with as well. For the most part (or perhaps entirely) it runs the same checks that happen in CI, so I'll look into clarifying that for new contributors. `make generate` is the most important part.

Can you tell me more about what failed when you tried to run it locally? `golangci-lint` has a history of taking unreasonably large amounts of both time and memory to run. So far it's been running for 40 minutes on my machine and I'm not yet sure if it will finish or fail.

mbbush:
Another option would be to submit a PR upstream to hashicorp/terraform-provider-aws to change the description, if it's wrong in the terraform context as well. If you do that, the description will update here in the crossplane upjet provider the next time we upgrade our upstream version.

apnsngr:
Thank you so much for all the information. It looks like the documentation issue is in the upstream Terraform provider. I thought I had checked this, but I must have been looking in the wrong spot. Terraform prefixes this description with "(Optional)" in their docs site which is a little more clear. I will look into making a PR to Terraform instead.

PR #1728, opened 2025-03-07
Update PR template to reference codegen

mbbush:

I just set up a clean dev environment, and while I was able to get `make reviewable` to run successfully, it took 55 minutes! It's possible it would have failed on a machine with less RAM available. This seems like an unreasonably high bar to expect from new contributors. I confess that I generally do not run it locally on my own PRs because it takes so long, and because most (all?) of its functionality is also implemented in CI. 

I think there's probably some more adjustments we can make to the language in the template to make this more clear, and I'm also interested in finding ways to improve the new-contributor experience as far as ability to run CI checks, although I understand there are some valid security concerns there. 

Fixes #

I have:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -20,7 +20,8 @@ Fixes #
 I have:
 
 - [ ] Read and followed Crossplane's [contribution process].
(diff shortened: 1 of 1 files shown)


jeanduplessis (approved):
This is a good step in the right direction to guide contributors.

turkenf (approved):
Thanks @mbbush 🙌

PR #1729, opened 2025-03-07
Tf api group overrides

mbbush:

When contributors add new resources, they are by default placed in an api group based on the second "word" in the terraform resource name. For most but not all resources, this matches the aws go sdk package name. 

We have an existing list of these exceptions, which I've updated to match the latest terraform provider version. By generating it programatically, I've found a few cases where we've introduced resource names which don't follow the convention, so I've added those as exceptions to avoid breaking changes.

I originally opened this PR a year ago as #1200, but it never reached the top of anyone's priority list. Tracking this will help make adding new resources easier. 

The scripts for parsing lists of terraform resources by framework/sdk will be useful for easing the burden of upstream provider upgrades. They're hacky and brittle, but they work, and nothing depends on them or runs them automatically, so they'll be easy to improve/replace if and when necessary. 

I have:

The changed codepaths are only used in code generation, and there is zero diff to the generated code. 


sergenyalcin:
@mbbush Thanks for this PR. If I understand correctly, these scripts are not included in the automation and generation pipelines. So they are like some helper tools. Although they are a bit fragile, if this is the case, then it makes sense to me. Also, it seems that they don't cause any difference in the generation now.

mbbush:
@sergenyalcin Correct on all points. The goal of the changes in `groups.go` are for futureproofing the addition of new resources. One of the main advantages of tracking the resource lists in source control is being able to see changes when we bump the terraform provider version.

PR #1730, opened 2025-03-10
Decouple config dependency reg org and publishing reg org

sergenyalcin:

This PR decouples config dependency reg org and publishing reg org. For more details please see here: https://github.com/crossplane/crossplane/issues/6320

There is dependency resolution issue for now since the dependencies shows the `ghcr.io`. With this change, there will be two different sources for dependency reg org and publishing reg org.

By default, they will be same but this change will make separately configurable them.

I have:


Makefile
```diff
@@ -130,9 +130,11 @@ export XPKG_IGNORE := $(XPKG_IGNORE)
 
 CONFIG_CRD_GROUP = $(PROVIDER_NAME)
 PROVIDER_AUTH_GROUP = $(PROVIDER_NAME)
+CONFIG_DEPENDENCY_REG_ORG = $(XPKG_REG_ORGS)
```
jastang, line 133:
```suggestion
CONFIG_DEPENDENCY_REG_ORG ?= $(XPKG_REG_ORGS)
```

what if we default to the push location only if not explicitly set (e.g. via https://github.com/crossplane-contrib/provider-workflows/pull/8) ?

jastang (approved):
lgtm; verified locally with a personal ghcr repository as well

Makefile
mbbush: I like the `?=` suggestion.

What stands out to me here is that we're setting a singular variable from a plural one. I have only a very vague recollection of why `XPKG_REG_ORGS` is plural; maybe there's support for publishing to multiple registries?

Ideally, we'd either 
1. Set the `CONFIG_DEPENDENCY_REG_ORG` to the first element of `XPKG_REG_ORGS`, (although it's not clear to me what the delimiter is)
2. Find/Replace XPKG_REG_ORGS (and perhaps also REGISTRY_ORGS?? with XPKG_REG_ORG if it actually only supports one.

mbbush (approved):
LGTM. The suggestion about plural/singular would be a nice improvement, but is optional.

Makefile
sergenyalcin: @mbbush While doing this change, it also attracted my attention. Although it is plural as a name, for example, it has a singular usage in the [templating](https://github.com/crossplane-contrib/provider-upjet-aws/blob/45ec02187ac6371229b719fc3eb277914a666552/cluster/images/provider-aws/Makefile#L82). Therefore, I will leave it like this for now, but this is a very good point. Thanks @mbbush.

sergenyalcin: Btw, @mbbush, @jastang, I will change the `=` -> `?=`

PR #1735, opened 2025-03-12
Consume terraform-provider-aws as a (patched) git submodule, not a fork

mbbush:

In an effort to make the (slight) differences between the upstream [hashicorp/terraform-provider-aws](https://github.com/hashicorp/terraform-provider-aws) and the [custom fork](https://github.com/upbound/terraform-provider-aws) this project has used since we started invoking the provider's go api more visible and easier to maintain, I've added the upstream provider as a git submodule, and a directory containing the patches the crossplane provider needs.

I'm very deliberately not including *any* behavior changes in this PR, only build process changes, so the resulting artifact should behave identically to the current one. I want to use this property as a way to validate that it is correct, but I'd appreciate some help with that. In particular, I don't have an arm64 machine I can use to test the local development workflow on. 

I'm borrowing heavily here from the pulumi project, which maintains patched forks of many terraform providers using a lot of Apache-2.0-licensed tooling. 

New makefile targets:

* `make upstream` Resets the `upstream` submodule to the ref committed to the parent project, then applies all the patches in order. Results in a detached head on the submodule containing exactly the source code we use. 
* `make upstream.build` Compile the patched upstream terraform provider from source. This is only really necessary because we use the terraform CLI to generate the json schema for the provider, rather than using the go schema, as a leftover from when we first moved to the "no-fork" architecture but needed to avoid some schema changes. I'd love to revisit this at some point, so that at least for new resources we use the go schema. 

I've adopted the "sentinel file" approach for tracking when a target needs to be rebuilt for the upstream patching target, added it as a dependency to all the make targets that seem to need it, and updated the check-diff target to pass when there are uncommitted changes to the upstream submodule. This results in a warning because I'm redefining a target from common.mk, which I plan to resolve by updating check-diff in the build submodule, but I want to wait just a little bit longer for the dust to settle before I do that.

There's a lot left to automate here, and lots of room for improvement. I welcome suggestions, and I'd like to learn more about what tooling @mergenci and @turkenf have already created for partially automating the version bumps. 

Changes:
* Since there's no longer an upstream repository to download release artifacts from, the schema generation now relies on building the terraform provider from source. I had hoped to find the `schema.json` somewhere else in the terraform provider, but it looks like their makefile generates it in a very similar way to ours. I adopted some patterns from theirs as it seemed much simpler. 

* Since we have the upstream repo in a submodule, we can just use its documentation directory to generate examples and references rather than pulling it from github.

The generated code is unchanged. I still need to come up with a good testing plan. If someone has examples of tests that failed without the terraform provider patches, those would be good candidates to run. Here are a few tests; I'm sure we'll want some more.

- [ ] Someone (not @mbbush) has tested the local development workflow on this branch using an arm64 mac, and it works. Matt only has access to amd64 linux boxes. 
- [ ] Verification that the multi-arch build process used for publishing still works correctly.


Makefile
```diff
@@ -194,32 +193,64 @@ $(TERRAFORM):
...
+	@if git status --porcelain build | grep . ; then $(ERR) There are uncommitted changes in the build submodule. Please ensure the build submodule is checked out at the desired commit. && false; else $(OK) build submodule is clean; fi
+	@$(INFO) checking that the upstream submodule
+	@if git status --porcelain --ignore-submodule=dirty | grep . ; then $(ERR) There are new commits in the upstream submodule. Please ensure all changes are written to patch files using scripts/upstream.sh. && false; else $(OK) upstream submodule is clean; fi
+
+
+.PHONY: upstream upstream.build
+
+generate.init: $(UPSTREAM) $(TERRAFORM_PROVIDER) $(TERRAFORM_PROVIDER_SCHEMA)
+
+build.init: $(UPSTREAM)
```
sergenyalcin, line 246: I think we are considering this will affect all build phases/targets of the provider?

Makefile
```diff
@@ -281,7 +312,7 @@ providerconfig-e2e-nopublish:
 uptest-local:
 	@$(WARN) "this target is deprecated, please use 'make uptest' instead"
 
-build-provider.%:
+build-provider.%: $(UPSTREAM)
```
sergenyalcin, line 315: Is that really necessary? Because we already put something in for `build.init`. I don't mind leaving it like this. I'm just trying to make sure I understand it correctly.

patches/0001-Copy-xpprovider-xpprovider.go-internal-conns-awsclie.patch
sergenyalcin, line 1: From this I understand that upstream modifications will now be made with the apply of these patches. In this context, do you think it would make sense to document how to get these files to help contributors?

Makefile
```diff
@@ -194,32 +193,64 @@ $(TERRAFORM):
...
+	@touch $@
+	@$(OK) Patched upstream terraform provider source
+
+# Apply patches to upstream tf provider source
+upstream: $(UPSTREAM)
+
+# Alias to build the upstream terraform provider
+upstream.build: $(TERRAFORM_PROVIDER)
+
+# Overwrite the check-diff target from the build submodule to exclude the upstream submodule.
```
sergenyalcin, line 231: I am a bit confused about this comment. We want to exclude the diff on upsteram submodules and already --`ignore-submodule=all` does this, right? Actually what we seem to be doing is adding to see if there is also diff in submodules, am I wrong? Is this really necessary? And if so, would you consider editing this comment for being more clear?

Makefile
```diff
@@ -194,32 +193,64 @@ $(TERRAFORM):
...
-	@git -C "$(WORK_DIR)/$(notdir $(TERRAFORM_PROVIDER_REPO))" sparse-checkout set "$(TERRAFORM_DOCS_PATH)"
+# Add sentinel files for tracking which targets need to be rebuilt
+UPSTREAM := $(ROOT_DIR)/.upstream.sentinel
 
-generate.init: $(TERRAFORM_PROVIDER_SCHEMA) pull-docs
+# Applies all the patches to the upstream files, but does not commit them.
+# Rebuild if the patches change, or upstream moves to a new HEAD, with some extra logic in the shell script
+$(UPSTREAM): $(wildcard patches/*) $(shell ./scripts/upstream.sh file_target)
+	@$(INFO) Patching upstream terraform provider source
+	@$(ROOT_DIR)/scripts/upstream.sh init -f
```
sergenyalcin, line 221: Just out of curiosity, is it really necessary to use `-f` here? Could it be that in the local environment the user has sometimes edited the upstream and is trying to test it? Is it right to overwrite it at this point?

Makefile
```diff
@@ -180,9 +180,8 @@ build.init: $(UP)
 # Setup Terraform for fetching provider schema
 TERRAFORM := $(TOOLS_HOST_DIR)/terraform-$(TERRAFORM_VERSION)
 TERRAFORM_WORKDIR := $(WORK_DIR)/terraform
-TERRAFORM_CLI_CONFIG_FILE := $(TERRAFORM_WORKDIR)/.terraformrc-custom
-TERRAFORM_FILESYSTEM_MIRROR := $(TOOLS_HOST_DIR)/$(TERRAFORM_PROVIDER_RELEASE)
-TERRAFORM_PROVIDER := $(TERRAFORM_FILESYSTEM_MIRROR)/registry.terraform.io/$(TERRAFORM_PROVIDER_SOURCE)/$(TERRAFORM_PROVIDER_VERSION)/$(SAFEHOST_PLATFORM)/terraform-provider-aws
+# The terraform aws provider hardcodes version 99.99.99 for local builds
+TERRAFORM_PROVIDER := $(ROOT_DIR)/upstream/terraform-plugin-dir/registry.terraform.io/$(TERRAFORM_PROVIDER_SOURCE)/99.99.99/$(SAFEHOST_PLATFORM)/terraform-provider-aws
```
sergenyalcin, line 184: I think still preserving the $(TERRAFORM_PROVIDER_VERSION) here makes more sense to me. Especially in terms of keeping multiple versions of the build in the local environment and making it look consistent.

But here I have a question let's consider there are more than one binary for different versions after my suggested change, then will the `-plugin-dir` flag confuse to use which binary?

Makefile
```diff
@@ -194,32 +193,64 @@ $(TERRAFORM):
...
+	@$(INFO) generating provider schema for $(TERRAFORM_PROVIDER_SOURCE) from the upstream git submodule
 	@rm -fr $(TERRAFORM_WORKDIR)
 	@mkdir -p $(TERRAFORM_WORKDIR)
-	@echo '{"terraform":[{"required_providers":[{"provider":{"source":"'"$(TERRAFORM_PROVIDER_SOURCE)"'","version":"'"$(TERRAFORM_PROVIDER_VERSION)"'"}}],"required_version":"'"$(TERRAFORM_VERSION)"'"}]}' > $(TERRAFORM_WORKDIR)/main.tf.json
-	@echo 'provider_installation { filesystem_mirror { path = "'"$(TERRAFORM_FILESYSTEM_MIRROR)"'", include = ["hashicorp/aws"] }, direct { exclude = ["hashicorp/aws"] } }' > $(TERRAFORM_CLI_CONFIG_FILE)
-	@TF_CLI_CONFIG_FILE=$(TERRAFORM_CLI_CONFIG_FILE) $(TERRAFORM) -chdir=$(TERRAFORM_WORKDIR) init -upgrade > $(TERRAFORM_WORKDIR)/terraform-logs.txt 2>&1
-	@TF_CLI_CONFIG_FILE=$(TERRAFORM_CLI_CONFIG_FILE) $(TERRAFORM) -chdir=$(TERRAFORM_WORKDIR) providers schema -json=true > $(TERRAFORM_PROVIDER_SCHEMA) 2>> $(TERRAFORM_WORKDIR)/terraform-logs.txt
-	@$(OK) generating provider schema for $(TERRAFORM_PROVIDER_SOURCE) $(TERRAFORM_PROVIDER_VERSION)
+	@echo 'data "aws_partition" "example" {}' > $(TERRAFORM_WORKDIR)/main.tf
```
sergenyalcin, line 209: Is this arbitrary? Why don't we use the old one? When we consider together with [the comment](https://github.com/crossplane-contrib/provider-upjet-aws/pull/1735/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R184), then will versioning work?

Makefile
```diff
@@ -194,32 +193,64 @@ $(TERRAFORM):
...
+$(TERRAFORM_PROVIDER): $(UPSTREAM)
+	@$(INFO) Building terraform AWS provider from source if changes detected
+	@(cd $(ROOT_DIR)/upstream && make go-build)
+	@$(OK) Built terraform AWS provider from source
 
+# TODO(mbbush) Remove this when we can. This is the last thing left over from when we used to fork a terraform cli
+# process for each resource. This generates the tf provider's json schema, which is slightly different from the go
+# schema (float vs int for some types). To avoid many schema changes, we added code in config/registry.go to use the
+# json schema for schema generation. Now that we have multi-version CRDs, it's easier to manage the schema change.
 $(TERRAFORM_PROVIDER_SCHEMA): $(TERRAFORM) $(TERRAFORM_PROVIDER)
```
sergenyalcin, line 205: Do you think should we add $(UPSTREAM) as a prerequisite for this target? Let's consider there is a change in the Go schema in the upstream then this will affect the generated json schema.

sergenyalcin (commented):
Thanks @mbbush for this PR. I left a few comments for discussion and your consideration.

Makefile
mbbush: Yes, my goal was to have the upstream target run before any of the other build targets. If you know of a better way to do that than specifying them all, I'd love to hear it.

Makefile
mbbush: I think you're right, it's probably unnecessary. I probably wrote this change before I figured out the build.init one, and just forgot to remove it.

patches/0001-Copy-xpprovider-xpprovider.go-internal-conns-awsclie.patch
mbbush: Yes, I can write something up for one of the .md files. The upstream.sh file also has a pretty good help interface.

Makefile
mbbush: It's already a transitive dependency via $(TERRAFORM_PROVIDER), so I didn't think it was necessary to include it here. I'd be happy to do so if you think it would improve readability, or if there's some subtle edge case about make files that requires it.

Makefile
mbbush: The challenge with that approach is patching the terraform provider's own build process to write the binary to a path that contains the desired version. I suppose we could also do it by copying the artifact to a path that we choose in the work dir, not the upstream submodule, but I found the simplicity of this approach appealing. 

Conceptually, we aren't using any published version of the tf provider any more, but rather one built from patched source code. So the hard coded 99.99.99 seemed fine to me. 

When working on this and the provider version upgrade, I was fairly frequently switching back and forth between the two tf provider versions, and rebuilding the tf provider. It was faster than I expected.

Makefile
```diff
@@ -194,32 +193,64 @@ $(TERRAFORM):
...
-	@echo 'provider_installation { filesystem_mirror { path = "'"$(TERRAFORM_FILESYSTEM_MIRROR)"'", include = ["hashicorp/aws"] }, direct { exclude = ["hashicorp/aws"] } }' > $(TERRAFORM_CLI_CONFIG_FILE)
-	@TF_CLI_CONFIG_FILE=$(TERRAFORM_CLI_CONFIG_FILE) $(TERRAFORM) -chdir=$(TERRAFORM_WORKDIR) init -upgrade > $(TERRAFORM_WORKDIR)/terraform-logs.txt 2>&1
-	@TF_CLI_CONFIG_FILE=$(TERRAFORM_CLI_CONFIG_FILE) $(TERRAFORM) -chdir=$(TERRAFORM_WORKDIR) providers schema -json=true > $(TERRAFORM_PROVIDER_SCHEMA) 2>> $(TERRAFORM_WORKDIR)/terraform-logs.txt
-	@$(OK) generating provider schema for $(TERRAFORM_PROVIDER_SOURCE) $(TERRAFORM_PROVIDER_VERSION)
+	@echo 'data "aws_partition" "example" {}' > $(TERRAFORM_WORKDIR)/main.tf
+	@$(TERRAFORM) -chdir=$(TERRAFORM_WORKDIR) init -plugin-dir $(ROOT_DIR)/upstream/terraform-plugin-dir > $(TERRAFORM_WORKDIR)/terraform-logs.txt 2>&1
+	@$(TERRAFORM) -chdir=$(TERRAFORM_WORKDIR) providers schema -json=true > $(TERRAFORM_PROVIDER_SCHEMA) 2>> $(TERRAFORM_WORKDIR)/terraform-logs.txt
+	@$(OK) generated provider schema for $(TERRAFORM_PROVIDER_SOURCE) from the upstream git submodule
```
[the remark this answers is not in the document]
mbbush: Yes, exactly!

Makefile
```diff
@@ -194,32 +193,64 @@ $(TERRAFORM):
...
-	@unzip $(TERRAFORM_PROVIDER).zip -d $(dir $(TERRAFORM_PROVIDER))
-	@rm -fr $(TERRAFORM_PROVIDER).zip
-	@$(OK) installing terraform provider $(SAFEHOST_PLATFORM)
+$(TERRAFORM_PROVIDER): $(UPSTREAM)
+	@$(INFO) Building terraform AWS provider from source if changes detected
+	@(cd $(ROOT_DIR)/upstream && make go-build)
+	@$(OK) Built terraform AWS provider from source
 
+# TODO(mbbush) Remove this when we can. This is the last thing left over from when we used to fork a terraform cli
+# process for each resource. This generates the tf provider's json schema, which is slightly different from the go
```
mbbush, line 202: As I've dug deeper into the various schemas, I'm less sure that we should remove this entirely, so I'll update the comment.

Makefile
mbbush: I'm not 100% certain if it's necessary. The big-picture idea of the workflow for making changes to the upstream provider is:

```
# Convert the patches in the patch directory into commits, and check out the 
# detached head created by applying them on top of the commit in the parent repo
./scripts/upstream.sh checkout

# change the code as needed
cd upstream
...

# commit your changes, and make/amend existing commits as needed
git commit --whatever

# convert the git commits back into patches, overwriting what's currently in the patches directory
cd ..
./scripts/upstream.sh check_in

# test your changes
...

# Commit your changes to the patch files to provider-upjet-aws
git commit -s -m "Changed the terraform provider to make it glow in the dark"
```

Makefile
mbbush: My goal was to make the smallest possible behavior change, keeping `check-diff` as strict as possible. 

Currently it fails if there's any change to any tracked file, any untracked files, or any submodule.

My idea is to update the job in the build submodule to ignore all submodules, but then additionally check the build submodule (which it knows shouldn't have changes), leaving other submodule checks to steps defined in the providers. For provider-upjet-aws, I expect that the submodule will be dirty, but still be at the same git reference committed to the parent repo, because you should have converted any local changes back into patch files using `scripts/upstream.sh check_in`. If not, it should fail.

I can definitely clarify some of the comments.

Makefile
mbbush: I added essentially that same thing :point_up:  to the README.md

Makefile
mbbush: I basically replaced the process we had for generating the schema.json file with the one the terraform provider itself had for generating the same file. I didn't want to try to implement a new build process for the upstream provider, so I was trying to keep it simple. 

As I think about it again, it seems like it might be possible for my simple workflow to download a provider from the terraform registry, which is not what we want, so I think I'll incorporate more of what was there before. To the extent I can, I'll avoid aws-specific things, so that as much as possible can be reused with the other providers. Thanks for the callout.

Makefile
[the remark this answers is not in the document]
mbbush: After thinking about it some more, I think the old method is perhaps a bit safer, so I'll change some of this back.

Makefile
mbbush: Also important to note: this path is precisely where `make go-build` in the terraform provider puts the compiled binary.

mbbush:
Thanks for the review @sergenyalcin!

I responded to your comments, and I think you're right that the existing process for specifying where to find the terraform provider binary is potentially safer than the one I implemented, so I'll make that change when I get a chance. I think supporting multiple terraform provider versions locally will actually be quite tricky, so I'd prefer sticking with the current v99.99.99. If you've got ideas about a better way to manage that I'd love to hear them.

I'm going to be on vacation for the next week, and away from my computer, so apologies in advance if I'm slow to respond.

Makefile
mbbush: But I think my preference is to leave it here for clarity.

mbbush:
/test-examples="examples/elbv2/v1beta2/lblistener-forward-single-targetgroup.yaml"

PR #1767, opened 2025-04-07
Configure and add an example for vpclattice service

turkenf:

This PR adds the [aws_vpclattice_service](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/vpclattice_service) resource.

I have:

Manually and uptest


turkenf:
/test-examples="examples/vpclattice/v1beta1/service.yaml"

package/crds/vpclattice.aws.upbound.io_services.yaml
```diff
@@ -0,0 +1,432 @@
...
+                - Delete
+                type: string
+              forProvider:
+                properties:
+                  authType:
+                    description: Type of IAM policy. Either NONE or AWS_IAM.
+                    type: string
+                  certificateArn:
+                    description: Amazon Resource Name (ARN) of the certificate.
+                    type: string
```
mbbush, line 80: This seems like a good candidate for a cross-resource reference

mbbush (approved):
Thanks @turkenf for the new resource!

One thing I haven't checked is whether the API group overrides are up to date for this resource.  I've got an [open PR](https://github.com/crossplane-contrib/provider-upjet-aws/pull/1729) to update them for all resources in the tf provider that would be great to have your review on, that would make the codegen do the right thing for newly added resources.

package/crds/vpclattice.aws.upbound.io_services.yaml
turkenf: Yes, this is the initial PR for the vpclattice API. We plan to add other resources in this service in the coming days, so we can handle the reference then.

turkenf:
Thanks for your review and PR @mbbush, I will check it soon. We intentionally did not add vpclattice to the ec2 group; the ec2 provider is already quite large.

PR #1780, opened 2025-05-07
Add Observe only CloudFormation Stack

tdeheurles:

Add an example of Cloudformation Stack Observe which behave in a specific way

I have:

- [X] Read and followed Crossplane's [contribution process].
- [ ] Run `make generate` and committed the results (ideally in a separate commit). 
- [ ] Not made any manual changes to generated files, and verified this with `make check-diff`.
--> I only updated documentation

This is documentation

### DIFF
--- a/examples/cloudformation/v1beta1/stack.yaml
+++ b/examples/cloudformation/v1beta1/stack.yaml
@@ -38,3 +38,18 @@ spec:
           }
         }
       }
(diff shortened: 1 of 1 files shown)


jeanduplessis:
@tdeheurles could you please look at signing off your commit so it can pass the DCO check?

Issue #1811, opened 2025-06-27
[Bug]: LBListener: Terraform validation fails during reconciliation when using `targetGroupArnSelector`


aswnrana:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Affected Resource(s)

"lblisteners.elbv2.aws.upbound.io/v1beta1"
"lblisteners.elbv2.aws.upbound.io/v1beta2"

### Description
This issue covers scenarios where managed resources get stuck in continuous reconciliation loops after successful initial creation via composition. The provider repeatedly attempts to update the resource but fails validation, even though the resource exists and is functional in the cloud provider. This typically occurs when there's a mismatch between how the provider handles initial creation versus subsequent reconciliation of resources with reference selectors.

## Composition manifest

[composition.txt](https://github.com/user-attachments/files/20944934/composition.txt) (File is too large to directly paste here, download this and change the extension to `.yaml`)

## Functions
```
apiVersion: pkg.crossplane.io/v1beta1
kind: Function
metadata:
  name: function-go-templating
spec:
  package: xpkg.upbound.io/upbound/function-go-templating:v0.10.0
---
apiVersion: pkg.crossplane.io/v1beta1
kind: Function
metadata:
  name: function-environment-configs
spec:
  package: xpkg.upbound.io/crossplane-contrib/function-environment-configs:v0.3.0

```

## Managed Resource
 *Directly provisioned resources are not causing issues*
```
apiVersion: elbv2.aws.upbound.io/v1beta2
kind: LBListener
metadata:
  labels:
    listenerDnat: mr-debug
  name: mr-debug-test
spec:
  forProvider:
    defaultAction:
    - targetGroupArnSelector:
        matchLabels:
          targetGroupDnat: mr-debug-test
      type: forward
    loadBalancerArnSelector:
      matchLabels:
        dnatNlbIb: fancy-aws-network
      policy:
        resolution: Required
        resolve: Always
    port: 80
    protocol: TCP
    region: eu-central-1
---
apiVersion: elbv2.aws.upbound.io/v1beta2
kind: LBTargetGroup
metadata:
  annotations:
  labels:
    targetGroupDnat: mr-debug-test
  name: test
spec:
  forProvider:
    name: mr-debug-test
    port: 80
    protocol: TCP
    region: eu-central-1
    targetType: ip
    vpcIdSelector:
      matchLabels:
        networks.aws.platform.upbound.io/network-id: configuration-aws-network

```

[...]

PR #1822, opened 2025-07-25
Configure SSA merge strategy for LBListener's spec.forProvider.defaultAction

ulucinar:

Related to: #1811

While investigating #1811, we've figured out that the SSA merge strategy for the `spec.forProvider.defaultAction` of the `LBListener.elbv2` MR is not configured properly. This PR proposes to inject a field named `index` into `spec.forProvider.defaultAction` with a default value of `default`. This will fix the merge behavior when there's only one item in the `spec.forProvider.defaultAction` list.

When specifying a single rule in `spec.forProvider.defaultAction`, one does not need to set the `index` field explicitly as it's defaulted. So the following works:
```yaml
apiVersion: elbv2.aws.upbound.io/v1beta2
kind: LBListener
metadata:
  name: listener
spec:
  forProvider:
    defaultAction:
    - type: forward
       targetGroupArnSelector:
         matchControllerRef: true
         policy:
           resolution: Required
           resolve: Always
       ...
```

This makes the change backwards-compatible.

If multiple rule items are to be specified in `spec.forProvider.defaultAction` and the user needs to configure the merge behavior of SSA, then she would need to manually configure the `index` fields as follows:
```yaml
apiVersion: elbv2.aws.upbound.io/v1beta2
kind: LBListener
metadata:
  name: listener
spec:
  forProvider:
    defaultAction:
    - index: "rule1"
       type: forward
       targetGroupArnSelector:
         matchControllerRef: true
         policy:
           resolution: Required
           resolve: Always
    - index: "rule2"
       type: ...
    ...
```

This allows to make the kube API treat the `spec.forProvider.defaultAction` as a list of type `map` and uses the list map key `index` when merging the list items.

We cannot use the `type` field as the map key and need to inject a new map key because the AWS API allows multiple rules of the same `type`. `order` is not an identifier field for the rules, i.e., `order` cannot be used to infer whether two items in the list represent the same item. For example, the orders of two rules might be interchanged, or a new element might be inserted before others in an API call. 

I have:

- [ x] Read and followed Crossplane's [contribution process].

Tested using a Crossplane configuration that patches the `spec.forProvider.defaultAction`.

### DIFF
--- a/apis/elbv2/v1beta1/zz_generated.deepcopy.go
+++ b/apis/elbv2/v1beta1/zz_generated.deepcopy.go
@@ -1933,6 +1933,11 @@ func (in *DefaultActionInitParameters) DeepCopyInto(out *DefaultActionInitParame
 			(*in)[i].DeepCopyInto(&(*out)[i])
 		}
 	}
+	if in.Index != nil {
+		in, out := &in.Index, &out.Index
+		*out = new(string)
+		**out = **in
+	}
 	if in.Order != nil {
 		in, out := &in.Order, &out.Order
 		*out = new(float64)
@@ -2008,6 +2013,11 @@ func (in *DefaultActionObservation) DeepCopyInto(out *DefaultActionObservation) 
 			(*in)[i].DeepCopyInto(&(*out)[i])
 		}
 	}
+	if in.Index != nil {
+		in, out := &in.Index, &out.Index
+		*out = new(string)
+		**out = **in
+	}
 	if in.Order != nil {
 		in, out := &in.Order, &out.Order
 		*out = new(float64)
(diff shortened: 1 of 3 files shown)


ulucinar:
/test-examples="examples/elbv2/v1beta2/lblistener.yaml"

ulucinar:
/test-examples="examples/elbv2/v1beta2/lblistener.yaml"

ulucinar:
/test-examples="examples/elbv2/v1beta2/lblistener.yaml"

turkenf (approved):
Thank you @ulucinar, LGTM 🙌

sergenyalcin (approved):
Thanks @ulucinar LGTM!

PR #1915, opened 2025-11-13
fix: use 12-digit fallback account ID

viren-nadkarni:

This PR makes the provider AWS client use a valid fallback account ID that has 12 digits. 

AWS account IDs always have [12 digits](https://docs.aws.amazon.com/accounts/latest/reference/manage-acct-identifiers.html). At the moment, the provider uses a 9 digit account ID which has a different meaning in [LocalStack configuration](https://docs.localstack.cloud/aws/capabilities/config/credentials/).

I have:

### DIFF
--- a/internal/clients/aws.go
+++ b/internal/clients/aws.go
@@ -30,7 +30,7 @@ const (
 	keyAccountID        = "account_id"
 	keyRegion           = "region"
 	keyPartition        = "partition"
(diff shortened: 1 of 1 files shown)


erhancagirici (approved):
Thanks @viren-nadkarni for fixing this! LGTM

PR #1965, opened 2026-02-14
fix(ecs): prevent panic in TerraformCustomDiff when taskDefinition is a plain family name

### LINKED ISSUE #1964: ECS Service TerraformCustomDiff panics when taskDefinition is a plain family name (v2.4.0)
## What happened?

The `TerraformCustomDiff` handler for `aws_ecs_service` (introduced in #1950) panics with an index-out-of-range error when `taskDefinition` is set to a plain family name (e.g., `"my-service"`) instead of an ARN.

The panic is silently caught by controller-runtime's global `recover()`, which prevents **all** drift detection on the resource. The `Synced` condition retains its previous `True` value — the resource appears healthy but is completely unmanaged. No error is reported, no events are emitted, and no reconciliation occurs.

## The bug

In `config/cluster/ecs/config.go`:

```go
// example td.New = arn:aws:ecs:us-west-1:<account_id>:task-definition/sampleservice:36
tdParts := strings.Split(td.New, "/")
if td.Old == tdParts[1] {  // panics when td.New has no "/"
    delete(diff.Attributes, "task_definition")
}
```

When `td.New` is a plain family name like `"my-service"`:
1. `strings.Split("my-service", "/")` returns `["my-service"]` (length 1)
2. `tdParts[1]` accesses index 1 of a length-1 slice → **panic: runtime error: index out of range [1] with length 1**

## Why it's silent

1. The panic propagates up from `TerraformCustomDiff` (called in `external_tfpluginsdk.go`)
2. controller-runtime's goroutine recover catches it
3. But the Synced condition is **not** updated — it keeps its previous value (True)
4. No error condition, no event, no log line from the reconciler
5. The provider pod shows 0 restarts and near-zero log activity

This makes the failure extremely difficult to diagnose, because every diagnostic surface (`kubectl get`, conditions, events) reports the resource as healthy.

## Impact

ALL drift detection on the affected `Service` managed resource is blocked — not just `task_definition`, but every field including tags. Any ECS Service using a plain family name for `taskDefinition` in v2.4.0 will silently stop receiving updates.

## How to reproduce

1. Create an `ecs.aws.upbound.io/v1beta2` `Service` resource
2. Set `spec.forProvider.taskDefinition` to a plain family name (e.g., `"my-service"`) rather than an ARN
3. Make any change to the Service spec (e.g., modify a tag)
4. Observe that the provider never reconciles the change — `Synced` stays `True` indefinitely

## Suggested fix

Add a bounds check before accessing `tdParts[1]`:

```go
tdParts := strings.Split(td.New, "/")
if len(tdParts) >= 2 && td.Old == tdParts[1] {
    delete(diff.Attributes, "task_definition")
}
```

When `taskDefinition` is a plain family name, the diff passes through unchanged and Terraform handles the comparison normally.

PR incoming: #XXXX

## Environment

- Provider version: v2.4.0
- Crossplane version: 1.18
- Upjet version: (bundled with provider)

/cc @ulucinar (author of #1950)

pyrex41:
## Description

Fixes #1964

The `TerraformCustomDiff` handler for `aws_ecs_service` (introduced in #1950) panics when `taskDefinition` is set to a plain family name instead of an ARN.

### The bug

```go
tdParts := strings.Split(td.New, "/")
if td.Old == tdParts[1] {  // panics when td.New has no "/"
```

When `td.New` is `"my-service"` (no `/`), `strings.Split` returns a single-element slice and `tdParts[1]` causes an index-out-of-range panic.

### Why it's severe

The panic is caught by controller-runtime's global recover, which:
- Does **not** update the `Synced` condition (it stays `True`)
- Does **not** emit any error events
- Silently blocks **all** drift detection on the resource (not just `task_definition` — all fields including tags)

The resource appears healthy but is completely unmanaged.

### The fix

Add a bounds check before accessing `tdParts[1]`:

```go
tdParts := strings.Split(td.New, "/")
if len(tdParts) >= 2 && td.Old == tdParts[1] {
    delete(diff.Attributes, "task_definition")
}
```

When `taskDefinition` is a plain family name, the diff passes through unchanged and Terraform handles the comparison normally.

### Context

The ECS API accepts `taskDefinition` as either:
- A full ARN: `arn:aws:ecs:us-east-2:123456:task-definition/my-service:5`
- A family name with revision: `my-service:5`
- A plain family name: `my-service` (resolves to latest ACTIVE revision)

The original code (from #1950) only handled the ARN case. This fix makes it safe for all three formats.

### DIFF
--- a/config/cluster/ecs/config.go
+++ b/config/cluster/ecs/config.go
@@ -83,9 +83,20 @@ func Configure(p *config.Provider) { //nolint:gocyclo
 			if td.Old == td.New {
 				return diff, nil
 			}
-			// example td.New = arn:aws:ecs:us-west-1:<account_id>:task-definition/sampleservice:36
+			// Suppress spurious diffs when the desired task_definition
+			// (without a revision) matches the family portion of the
+			// current (revision-qualified) value.
+			//
+			// td.New may be:
+			//   - An ARN: "arn:aws:ecs:us-west-1:123456:task-definition/svc:36"
+			//   - A plain family name: "svc"
+			//
+			// Only attempt the comparison when td.New contains "/"
+			// (i.e., is an ARN). A plain family name cannot be
+			// meaningfully compared to the revision-qualified td.Old,
+			// so we let the diff through and let Terraform handle it.
 			tdParts := strings.Split(td.New, "/")
-			if td.Old == tdParts[1] {
+			if len(tdParts) >= 2 && td.Old == tdParts[1] {
 				delete(diff.Attributes, "task_definition")
 			}
 			return diff, nil
--- a/config/namespaced/ecs/config.go
+++ b/config/namespaced/ecs/config.go
@@ -83,9 +83,20 @@ func Configure(p *config.Provider) { //nolint:gocyclo
 			if td.Old == td.New {
 				return diff, nil
 			}
-			// example td.New = arn:aws:ecs:us-west-1:<account_id>:task-definition/sampleservice:36
+			// Suppress spurious diffs when the desired task_definition
+			// (without a revision) matches the family portion of the
+			// current (revision-qualified) value.
+			//
+			// td.New may be:
+			//   - An ARN: "arn:aws:ecs:us-west-1:123456:task-definition/svc:36"
+			//   - A plain family name: "svc"
+			//
+			// Only attempt the comparison when td.New contains "/"
+			// (i.e., is an ARN). A plain family name cannot be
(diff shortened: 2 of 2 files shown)


stevendborrelli:
Hello @pyrex41 can you sign this commit? The lack of DCO is blocking further review. See <https://github.com/crossplane-contrib/provider-upjet-aws/pull/1965/checks?check_run_id=63597234437> on how to sign your commit.

PR #2082, opened 2026-05-18
fix: seed arn from id for aws_ecs_task_definition cold-start observe

### LINKED ISSUE #2080: [Bug]: ECS TaskDefinition Observe always fails with "external resource does not exist" on cold start
## Summary

`ecs.aws.m.upbound.io/v1beta1 TaskDefinition` resources in `Observe`-only management policy always reconcile with `SYNCED=False` and the message **\"external resource does not exist\"** on a cold start (i.e. when the managed resource has never been reconciled into a ready state in the current controller instance). The resource exists in AWS and the `crossplane.io/external-name` annotation is set correctly; the error is caused by a bug in how the underlying Terraform resource implementation looks up the task definition.

## Environment

| Field | Value |
|---|---|
| Provider | `provider-upjet-aws` (`github.com/upbound/provider-aws/v2`) |
| Upstream TF provider | `github.com/upbound/terraform-provider-aws` @ `f7691456b787` |
| upjet | `v2.2.1-0.20260414070754-c6d5213346ac` |
| API group | `ecs.aws.m.upbound.io/v1beta1` |
| Resource | `TaskDefinition` |

## Steps to Reproduce

1. Create a `TaskDefinition` managed resource with `managementPolicies: ["Observe"]` and `crossplane.io/external-name` set to a valid task definition ARN (e.g. `arn:aws:ecs:us-west-2:123456789012:task-definition/my-service:36`) or family name (`my-service`).
2. Apply to a cluster running this provider.
3. Observe the resource — it will never become `SYNCED=True`.

```yaml
apiVersion: ecs.aws.m.upbound.io/v1beta1
kind: TaskDefinition
metadata:
  name: my-service-taskdef
  annotations:
    crossplane.io/external-name: "arn:aws:ecs:us-west-2:123456789012:task-definition/my-service:36"
spec:
  managementPolicies: ["Observe"]
  forProvider:
    region: us-west-2
    family: my-service
    containerDefinitions: "[]"
```

**Result:**
```
SYNCED   False
Message: observe failed: cannot run refresh: refresh failed: "external resource does not exist"
```

The same behaviour is seen with all external-name formats: full ARN, `family:revision`, and family-only.

## Root Cause

The Terraform resource `aws_ecs_task_definition` uses [`d.Get("arn")`](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/ecs/task_definition.go) instead of `d.Id()` in `resourceTaskDefinitionRead` to construct the `DescribeTaskDefinition` API call.

In the upjet observe flow, `arn` is a **computed-only** attribute. On a cold start, upjet only populates `params["id"]` (from `crossplane.io/external-name`); `params["arn"]` is empty because it has not yet been read back from AWS. The Terraform read function therefore calls `DescribeTaskDefinition` with an empty identifier, which AWS rejects with a `400 InvalidParameter` — upjet interprets this as "resource not found" and surfaces it as `external resource does not exist`.

The `id` field (which equals the external-name) always has the correct value, but the read function never consults it.

## Proposed Fix

A `TerraformConfigurationInjector` on the `aws_ecs_task_definition` resource configurator seeds `params["arn"]` from `params["id"]` when `arn` is not yet populated:

```go
// config/cluster/ecs/config.go (and config/namespaced/ecs/config.go)
p.AddResourceConfigurator("aws_ecs_task_definition", func(r *config.Resource) {
    // ... existing References config ...

    // resourceTaskDefinitionRead uses d.Get("arn") instead of d.Id() to
    // call DescribeTaskDefinition. On a cold-start observe, "arn" is a
    // computed-only attribute and is not present in params, so the API
    // call goes out with an empty identifier and AWS returns a 400 that
    // surfaces as "external resource does not exist". Seed params["arn"]
    // from params["id"] (the external-name) so the read always has a
    // non-empty identifier regardless of which format was used.
    r.TerraformConfigurationInjector = func(_ map[string]any, params map[string]any) error {
        if arn, _ := params["arn"].(string); arn == "" {
            if id, ok := params["id"].(string); ok && id != "" {
                params["arn"] = id
            }
        }
        return nil
    }
})
```

This fix should be applied to both `config/cluster/ecs/config.go` and `config/namespaced/ecs/config.go`.

After the first successful observe, upjet will populate `params["arn"]` from the real AWS ARN returned in `atProvider`, so the injector becomes a no-op on subsequent reconciles.

## Integration Test Results

Tested in observe mode with provider-aws-ecs v2.5.3 against an existing ECS service in `us-west-2` after applying the `TerraformConfigurationInjector` fix locally:

| Resource | SYNCED | READY |
|---|---|---|
| Cluster | True | True |
| Service | True | True |
| TaskDefinition | **True** | **True** |

Without the fix, `TaskDefinition` stays `SYNCED=False` indefinitely across all supported external-name formats.

## Additional Context

This was discovered while building a Crossplane composition (`XECS`) that imports existing ECS stacks in observe mode. A workaround is available via `TerraformConfigurationInjector` in a fork, but the fix belongs upstream so all users of the provider benefit.

johncarpio-sumo:
## Summary

Fixes #2080

`ecs.aws.m.upbound.io/v1beta1 TaskDefinition` resources in `Observe`-only management policy always fail with `SYNCED=False` / `"external resource does not exist"` on cold start. The resource exists in AWS; the error comes from how the Terraform resource performs its read.

## Root Cause

`resourceTaskDefinitionRead` calls `d.Get("arn")` instead of `d.Id()` to build the `DescribeTaskDefinition` request. In the upjet observe flow, `arn` is computed-only — on cold start it is absent from `params` because it hasn't been read back from AWS yet. The API call goes out with an empty identifier, AWS returns `400 InvalidParameter`, and upjet surfaces that as "external resource does not exist".

The external-name annotation always has the correct value but is never consulted for this field.

## Fix

Override `SetIdentifierArgumentFn` on the `aws_ecs_task_definition` resource configurator (both `cluster` and `namespaced` variants) to seed `params["arn"]` from the external name when `arn` is not yet populated:

```go
r.ExternalName.SetIdentifierArgumentFn = func(base map[string]any, externalName string) {
    if arn, _ := base["arn"].(string); arn == "" && externalName != "" {
        base["arn"] = externalName
    }
}
```

This hook runs during the `Connect` phase (inside `getExtendedParameters`) before state reconstruction. After the first successful observe, upjet populates `params["arn"]` from the real AWS ARN in `atProvider`, so the function is a no-op on subsequent reconciles.

**Why `SetIdentifierArgumentFn` instead of `TerraformConfigurationInjector`?**
The ECS TaskDefinition controller uses `TerraformPluginSDKAsyncConnector`. During local testing we found that `TerraformConfigurationInjector` is not invoked in this async path, while `SetIdentifierArgumentFn` is reliably called during `getExtendedParameters`.

## Local End-to-End Testing

Tested by building the provider locally and running against a kind cluster with a real AWS ECS TaskDefinition in observe-only mode.

**Result:** Cold-start observe goes directly to `SYNCED=True READY=True` on first reconciliation.

### Steps to reproduce

```bash
# 1. Build the ECS provider binary
go build -o _output/bin/provider-ecs ./cmd/provider/ecs

# 2. Create a kind cluster with Crossplane
kind create cluster --name ecs-test
helm repo add crossplane-stable https://charts.crossplane.io/stable
helm install crossplane crossplane-stable/crossplane \
  --namespace crossplane-system --create-namespace --wait

# 3. Install required CRDs
kubectl apply -f package/crds/ecs.aws.m.upbound.io_taskdefinitions.yaml
kubectl apply -f package/crds/aws.m.upbound.io_clusterproviderconfigs.yaml
kubectl apply -f package/crds/aws.m.upbound.io_providerconfigusages.yaml

# 4. Create AWS credentials secret and ClusterProviderConfig
kubectl create secret generic aws-creds -n crossplane-system \
  --from-literal=credentials="[default]
aws_access_key_id = <YOUR_KEY>
aws_secret_access_key = <YOUR_SECRET>"

kubectl apply -f - <<YAML
apiVersion: aws.m.upbound.io/v1beta1
kind: ClusterProviderConfig
metadata:
  name: default
spec:
  credentials:
    source: Secret
    secretRef:
      namespace: crossplane-system
      name: aws-creds
      key: credentials
YAML

# 5. Create a real ECS TaskDefinition in AWS
aws ecs register-task-definition \
  --family cold-start-observe-test \
  --container-definitions '[{"name":"test","image":"alpine:latest","memory":128}]' \
  --requires-compatibilities FARGATE \
  --network-mode awsvpc --cpu "256" --memory "512"
# Note the ARN from the output

# 6. Run the provider locally
./_output/bin/provider-ecs --debug --certs-dir="" --poll=5s &

# 7. Apply an observe-only TaskDefinition MR
kubectl apply -f - <<YAML
apiVersion: ecs.aws.m.upbound.io/v1beta1
kind: TaskDefinition
metadata:
  name: cold-start-test
  annotations:
    crossplane.io/external-name: "arn:aws:ecs:us-east-1:<ACCOUNT>:task-definition/cold-start-observe-test:1"
spec:
  managementPolicies: ["Observe"]
  forProvider:
    region: us-east-1
  providerConfigRef:
    kind: ClusterProviderConfig
    name: default
YAML

# 8. Verify — should reach SYNCED=True READY=True within ~20s
kubectl get taskdefinition cold-start-test -o wide
```

### Expected result

```
NAME              SYNCED   READY   EXTERNAL-NAME             AGE
cold-start-test   True     True    cold-start-observe-test   20s
```

### Without the fix (on main branch)

```
NAME              SYNCED   READY   EXTERNAL-NAME                                                                  AGE
cold-start-test   False            arn:aws:ecs:us-east-1:417068142286:task-definition/cold-start-observe-test:1   3m
```

With error: `observe failed: external resource does not exist`

## Unit Tests

Added table-driven tests in both `config/cluster/ecs/config_test.go` and `config/namespaced/ecs/config_test.go` covering:
- Cold start with full ARN external name
- Cold start with family:revision format
- Warm start (arn already populated — no-op)
- Empty external name (no-op)

```bash
go test ./config/cluster/ecs/ ./config/namespaced/ecs/ -v
```

## Checklist

- [x] Fixes root cause without changing the observable API surface
- [x] Applied to both `config/cluster/ecs/config.go` and `config/namespaced/ecs/config.go`
- [x] Unit tests added for both variants
- [x] End-to-end tested locally in kind with real AWS resources
- [x] Idempotent and safe on subsequent reconciles


config/cluster/ecs/config.go
```diff
@@ -119,5 +119,17 @@ func Configure(p *config.Provider) { //nolint:gocyclo
...
 			},
 		}
+		// resourceTaskDefinitionRead uses d.Get("arn") instead of d.Id() to
+		// call DescribeTaskDefinition. On a cold-start observe, "arn" is a
+		// computed-only attribute and is not present in params, so the API
+		// call goes out with an empty identifier and AWS returns a 400 that
+		// surfaces as "external resource does not exist". Seed params["arn"]
+		// from the external name (via SetIdentifierArgumentFn) so the read
+		// always has a non-empty identifier regardless of which format was used.
+		r.ExternalName.SetIdentifierArgumentFn = func(base map[string]any, externalName string) {
```
erhancagirici: Could you move this implementation to `config/externalname.go`, for convention. It deduplicates the configuration (single config for cluster & namespaced), and makes it easier to get a proper overview of the external name configs.

config/cluster/ecs/config.go
```diff
@@ -119,5 +119,17 @@ func Configure(p *config.Provider) { //nolint:gocyclo
...
+			if arn, _ := base["arn"].(string); arn == "" && externalName != "" {
```
erhancagirici: no action needed here, just some notes on the change: 
- the `Read` implementation of the TF resource always sets the ID from the `family` parameter. This resource will switch its ID `arn` to `family` after initial read.
- TF `Read` is conditional on `track_latest` parameter. When true, It sends the `family` name in `DescribeTaskDefinition` AWS API call, otherwise the `arn`.

In practice, one can also import using the `family` parameter, like `my-task-def`, or `my-task-def:5` with revision, or the full ARN.
In that case, the external name is always carried to `arn` attribute during cold start, regardless of its shape.

Normally, we would've expected to parse the external-name and switch accordingly on which identifier to set. But for this particular case, this seems to be fine though, as the `arn` parameter is directly passed to `DescribeTaskDefinition` AWS SDK call, and accepts all shapes.

AWS SDK [`DescribeTaskDefinitionInput`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ecs@v1.72.1#DescribeTaskDefinitionInput) documentation reference for `TaskDefinition` field:
```
The family for the latest ACTIVE revision, family and revision ( family:revision ) for a specific revision in the family, or full Amazon Resource Name (ARN) of the task definition to describe.
```
And after the first observation, `arn` will take its proper shape 👍

config/namespaced/ecs/config.go
```diff
@@ -119,5 +119,17 @@ func Configure(p *config.Provider) { //nolint:gocyclo
...
 			},
 		}
+		// resourceTaskDefinitionRead uses d.Get("arn") instead of d.Id() to
+		// call DescribeTaskDefinition. On a cold-start observe, "arn" is a
+		// computed-only attribute and is not present in params, so the API
+		// call goes out with an empty identifier and AWS returns a 400 that
+		// surfaces as "external resource does not exist". Seed params["arn"]
+		// from the external name (via SetIdentifierArgumentFn) so the read
+		// always has a non-empty identifier regardless of which format was used.
+		r.ExternalName.SetIdentifierArgumentFn = func(base map[string]any, externalName string) {
```
erhancagirici: see https://github.com/crossplane-contrib/provider-upjet-aws/pull/2082/changes#r3383433002

erhancagirici (changes requested):
@johncarpio-sumo thanks for the contribution! The change looks good alreadty logic-wise, just requesting changes on the placement of the code, for the sake of convention. Also, please rebase the changes to latest main for more complete testing.

config/cluster/ecs/config.go
johncarpio-sumo: Done — moved to `config/externalname.go` as `ecsTaskDefinition()` function. Removed from both cluster and namespaced configs. Also rebased onto latest main.

config/cluster/ecs/config.go
johncarpio-sumo: Thanks for the detailed notes! Good to know all identifier shapes (ARN, family:revision, family) are accepted by DescribeTaskDefinition — confirms we don't need to parse/transform the external name before setting it.

config/namespaced/ecs/config.go
johncarpio-sumo: Addressed — removed from namespaced config, now lives in externalname.go.

johncarpio-sumo: @erhancagirici

config/cluster/ecs/config.go
johncarpio-sumo: @erhancagirici

erhancagirici:
/test-examples="examples/ecs/cluster/v1beta2/taskdefinition.yaml"

erhancagirici (approved):
@johncarpio-sumo thanks for the contribution! LGTM.
I've triggered the tests, after they are green we can merge.

One last thing, could you sign the CLA again? Looks like it was reset somehow.

johncarpio-sumo:
@erhancagirici here for the CLA let me know if I should do something else
Captura de pantalla 2026-06-10 a la(s) 3 11 44 p  m
Captura de pantalla 2026-06-10 a la(s) 3 11 48 p  m

erhancagirici:
/test-examples="examples/ecs/cluster/v1beta2/taskdefinition.yaml"

erhancagirici:
/test-examples="examples/ecs/cluster/v1beta2/taskdefinition.yaml"

PR #2096, opened 2026-06-03
Add ProviderConfig ReconciliationPolicy API

ulucinar:

Related PR: https://github.com/crossplane/upjet/pull/665

  This PR introduces a new reconciliationPolicy API on the ProviderConfig specs and tidies up the upjet controller-setup template.

  1. New spec.reconciliationPolicy.exponentialFailureRateLimiter API on `ProviderConfig`s

  Adds a new field on the cluster-scoped `ProviderConfig`, the namespaced `ProviderConfig`, and the `ClusterProviderConfig` that lets users tune the base-delay and the
  max-delay of the exponential-failure rate limiter applied to the managed-resource controllers. Previously these were hard-coded to `time.Second` / `60*time.Second` in every
  generated controller; they can be overridden at runtime from the resolved `ProviderConfig`.

  Hand-written changes for this commit:
  - `config/registry_cluster.go`, `config/registry_namespaced.go` — Wire up a project-local controller template (`config.WithControllerTemplate(templates.ControllerTemplate)`),
  required so the rate-limiter wiring (and subsequent template tweaks) can land in the generated controllers.
  - `config/templates/controller.go.tmpl`, `config/templates/embed.go` — new template + `go:embed` loader.
  - `internal/clients/pc_resolver.go` — new `ReconciliationPolicy(...)` helper that resolves a managed resource's effective `ProviderConfig` and returns its
  `Spec.ReconciliationPolicy`.
  - `ProviderConfig` CRDs (`package/crds/aws.upbound.io_providerconfigs.yaml`, `aws.m.upbound.io_providerconfigs.yaml`, `aws.m.upbound.io_clusterproviderconfigs.yaml`) — schema
  additions for the new field.

All `internal/controller/**/zz_controller.go` files regenerated to consume the per-MR rate limiter and the new finalizer/reconciler from `crossplane/upjet/v2/pkg/reconciler/reconciliationpolicy`.

An example `ProviderConfig`, through which you can override the exponential failure rate limiter parameters, looks like the following:
```yaml
kind: ProviderConfig
metadata:
  name: longer-backoff  # relative to the default max-delay of 1m
spec:
  credentials:
    secretRef:
      key: credentials
      name: example-aws-creds
      namespace: crossplane-system
    source: Secret
  reconciliationPolicy:
    exponentialFailureRateLimiter:
      baseDelay: 1s
      maxDelay:  5m
```
Managed resources (MRs) that use the above `ProviderConfig` will be associated with an exponential failure rate limiter whose base-delay is 1s and max-delay is 5m.

⚠️ **Note**: The provider currently does *not* allow a base-delay smaller than the default 1s and it does *not* allow a max-delay smaller than the default 60s. Also the max-delay must be greater than or equal to the configured base-delay. If, in a `ProviderConfig`, the base-delay is not specified (it's optional), then a base-delay of 1s is assumed. Similarly, if the max-delay is not specified (it's optional), then a max-delay of 60s is assumed.

**Note**: If there's no override for the exponential failure rate limiter parameters for an MR, its reconciler still uses the (previous) default parameters of a 1s base-delay and a 60s max-delay. The changes being introduced in this PR are backward-compatible.

⚠️ **Note**: If you update the `ProviderConfig`, we don't immediately queue reconciliation requests for the referencing MRs. This has been the behavior for the provider so far, and this PR does not change that behavior. There might be a huge number of MRs referencing a `ProviderConfig` and requeueing reconciliation requests for all them could have a similar effect to a provider pod restart. 

⚠️ **Note**: If you update the exponential failure rate limiter parameters in the `ProviderConfig`, the retry counters of the referencing MRs will be reset, their reconcilers will start backing-off with delays starting from the configured new base-delay with a cap at the configured max-delay. But as noted above, updating the `ProviderConfig` does not immediately queue a reconciliation request for the MR. The MR will be reconciled at its next scheduled time (if the MR has been in an error state, probably with an increased back-off delay according to the old (before the update) rate limiter parameters from the `ProviderConfig`, and if its reconciler encounters an error during the next reconciliation, the reconciler will back-off initially with the new base-delay, and if the error condition continues, will continue backing-off with exponentially growing delays, up to the new max-delay.
TLDR; changing the rate limiter parameters resets the back-off counters of the MRs referencing the updated `ProviderConfig`. Also, syntactical changes in the base-delay & max-delay that correspond to the same time durations, e.g., updating a `1s` to a `1000ms`, are *not* considered as changes, meaning that the retry counters (of the referencing MRs) will *not* be reset.

  2. Readability cleanup of the controller template

  A non-functional cleanup of `config/templates/controller.go.tmpl`. The generated `zz_controller.go` files are functionally identical (same opts contents, same effective configuration of `managed.NewReconciler` and `ctrl.NewControllerManagedBy`), but layout / hygiene is improved:
  - All conditional `opts = append(...)` blocks are now contiguous; `EnableAlphaChangeLogs` no longer sits past the webhook and state-metrics registration blocks.
  - Closing parentheses no longer pile up; each lands on its own line with the required trailing commas.
  - Import block in the template has `crossplane-runtime/v2/pkg/errors` in correct alphabetical position; the dead `crossplane/upjet/v2/pkg/terraform` import is removed.
  - ...

I have:

Manually tested the new API using the following provider packages:
- `xpkg.upbound.io/upbound/provider-family-aws:v2.6.0-d97ebe06bf0b`
- `xpkg.upbound.io/upbound/provider-aws-rds:v2.6.0-d97ebe06bf0b`

An MR, referencing the following `ProviderConfig` was put into a permanent error state:
```yaml
kind: ProviderConfig
metadata:
  name: longer-backoff
spec:
  credentials:
    secretRef:
      key: credentials
      name: example-aws-creds
      namespace: crossplane-system
    source: Secret
  reconciliationPolicy:
    exponentialFailureRateLimiter:
      baseDelay: 1s
      maxDelay: 5m0s
```


ulucinar:

ulucinar:

internal/clients/pc_resolver.go
```diff
@@ -143,3 +144,11 @@ func resolveProviderConfigModern(ctx context.Context, crClient client.Client, mg
 	}
 	return effectivePC, nil
 }
+
+func ReconciliationPolicy(ctx context.Context, client client.Client, mg xpresource.Managed) (*v1alpha1.ReconciliationPolicy, error) {
```
sergenyalcin, line 148: I am not sure where is the suitable place to comment for this but I preferred here:

If I understand correctly, `setRateLimiter` fetches the managed resource here, and then `managed.Reconciler` fetches it again. Same story on the `ProviderConfig` side: `clients.ReconciliationPolicy` reads it via `resolveProviderConfig`, and the connect phase reads it again for credential resolution. I believe both are cache reads through the caching client, so this may not be a real cost today.

The possible fix would be to stash the already-fetched object in the context so the inner reconciler can reuse it, but that requires touching crossplane-runtime's `managed.Reconciler` which is out of scope here. Just leaving a comment for future consideration.

apis/cluster/v1beta1/types.go
```diff
@@ -6,11 +6,16 @@ package v1beta1
...
 	xpv1 "github.com/crossplane/crossplane-runtime/v2/apis/common/v1"
+	"github.com/crossplane/upjet/v2/apis/configuration/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // A ProviderConfigSpec defines the desired state of a ProviderConfig.
 type ProviderConfigSpec struct {
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="!has(self.exponentialFailureRateLimiter) || !has(self.exponentialFailureRateLimiter.baseDelay) || has(self.exponentialFailureRateLimiter.maxDelay) || duration(self.exponentialFailureRateLimiter.baseDelay) <= duration('60s')",message="when maxDelay is omitted it defaults to 60s; baseDelay must be <= 60s"
+	ReconciliationPolicy *v1alpha1.ReconciliationPolicy `json:"reconciliationPolicy"`
```
sergenyalcin: Shouldn't there be an `omitempty` JSON tag here?

sergenyalcin (commented):
Thanks, @ulucinar. I've left two comments for discussion.

internal/clients/pc_resolver.go
ulucinar: Hi @sergenyalcin,
Thanks for bringing this up. Yes, correct, they are both reads from the shared informer caches and we should be fine.

apis/cluster/v1beta1/types.go
ulucinar: Done. Thanks for catching this! Added the `omitempty` tag option to both places.

ulucinar:

sergenyalcin (approved):
Thanks @ulucinar LGTM!

ulucinar:

PR #1975, opened 2026-02-26
Fix API Gateway v1 external name oscillation causing perpetual delete/recreate

aditmeno:
## Summary

Fixes #1974

7 API Gateway v1 resources using `FormattedIdentifierFromProvider("/", ...)` suffer from an external-name annotation oscillation bug. Each reconciliation cycle the `crossplane.io/external-name` annotation changes, triggering perpetual delete/recreate cycles and AWS 429 throttling.

### Root Cause

`FormattedIdentifierFromProvider` inherits `GetExternalNameFn = IDAsExternalName` (reads `tfstate["id"]` as-is) but overrides `GetIDFn` to join parameters with `/`. The Terraform AWS provider stores internal IDs in a **different format** than the import format:

| Resource | `GetIDFn` produces (slash) | `tfstate["id"]` stores (internal) |
|---|---|---|
| `aws_api_gateway_method` | `{restApiId}/{resourceId}/{httpMethod}` | `agm-{restApiId}-{resourceId}-{httpMethod}` |
| `aws_api_gateway_integration` | `{restApiId}/{resourceId}/{httpMethod}` | `agi-{restApiId}-{resourceId}-{httpMethod}` |
| `aws_api_gateway_stage` | `{restApiId}/{stageName}` | `ags-{restApiId}-{stageName}` |
| `aws_api_gateway_method_settings` | `{restApiId}/{stageName}/{methodPath}` | `{restApiId}-{stageName}-{methodPath}` |
| `aws_api_gateway_gateway_response` | `{restApiId}/{responseType}` | `aggr-{restApiId}-{responseType}` |
| `aws_api_gateway_integration_response` | `{restApiId}/{resourceId}/{httpMethod}/{statusCode}` | `agir-{restApiId}-{resourceId}-{httpMethod}-{statusCode}` |
| `aws_api_gateway_method_response` | `{restApiId}/{resourceId}/{httpMethod}/{statusCode}` | `agmr-{restApiId}-{resourceId}-{httpMethod}-{statusCode}` |

Since `GetExternalNameFn` returns the internal format (e.g., `agm-abc-def-GET`) but the controller expects the slash format (e.g., `abc/def/GET`), the annotation oscillates every reconciliation.

### Fix

New helper `apiGatewayFormattedIdentifier(prefix, keys...)` that bypasses `tfstate["id"]` entirely and reads individual tfstate fields directly:

[...]

PR #2117, opened 2026-06-12
Fix VPC endpoint association external name oscillation

### LINKED ISSUE #2116: VPC endpoint association resources have external name oscillation due to ID format mismatch
### What happened?

`VPCEndpointSubnetAssociation` (and `VPCEndpointRouteTableAssociation`) managed resources flip their `crossplane.io/external-name` annotation between two formats on every reconciliation, causing perpetual updates (and, on affected setups, delete/recreate churn and AWS `429` throttling).

Observed live across reconciles for a single subnet association MR:

```
ext=vpce-079f2245bf1ec40be/subnet-012d3109ebb773a1c    <-- import format
ext=a-vpce-079f2245bf1ec40be1284373945                 <-- internal-id format
```

The resource's `resourceVersion` increments continuously while it sits in the `vpce/subnet` state; a sibling association that happens to settle on the `a-<hash>` form is stable.

### Root cause

`aws_vpc_endpoint_subnet_association` and `aws_vpc_endpoint_route_table_association` are configured with `FormattedIdentifierFromProvider("/", ...)` in `config/externalname.go`. That helper leaves `GetExternalNameFn` as the `IdentifierFromProvider` default (reads `tfstate["id"]`) while `GetIDFn` joins the parameters with `/`.

But the Terraform AWS provider sets the id to a **prefixed, hashcode-based** value, not the slash-joined import id (`internal/service/ec2/id.go`):

```go
func vpcEndpointSubnetAssociationCreateID(vpcEndpointID, subnetID string) string {
    return fmt.Sprintf("a-%s%d", vpcEndpointID, create.StringHashcode(subnetID))
}
```

So `GetExternalNameFn` yields `a-<vpc_endpoint_id><hashcode(subnet_id)>` while `GetIDFn` yields `vpc_endpoint_id/subnet_id` — the two disagree and the external name oscillates. This is the same class of bug fixed for the API Gateway v1 resources in #1974 / #1975.

(`aws_vpc_endpoint_security_group_association` uses `config.IdentifierFromProvider` — both sides are the `a-<hash>` id, so it is consistent and unaffected.)

### How can we reproduce it?

Create a `VPCEndpointSubnetAssociation` and watch `kubectl get vpcendpointsubnetassociations -o yaml` — the `crossplane.io/external-name` annotation alternates between `<vpc_endpoint_id>/<subnet_id>` and `a-<vpc_endpoint_id><number>`, and `resourceVersion` climbs continuously.

### Fix

Make `GetExternalNameFn` and `GetIDFn` both deterministic from the named fields (same approach as #1975): external name = `vpc_endpoint_id/child_id`, Terraform id = `a-<vpc_endpoint_id><hashcode(child_id)>`. PR to follow.

aditmeno:

`aws_vpc_endpoint_subnet_association` and `aws_vpc_endpoint_route_table_association` were configured with `FormattedIdentifierFromProvider("/", ...)`, which has the same external-name oscillation bug that the API Gateway v1 resources had (fixed in #1975):

- `GetExternalNameFn` falls back to `IdentifierFromProvider`'s default and reads `tfstate["id"]`, which the AWS provider sets to `a-<vpc_endpoint_id><hashcode(child_id)>` (`internal/service/ec2.vpcEndpoint*AssociationCreateID`);
- `GetIDFn` joins the parameters with `/` to produce the documented import id `vpc_endpoint_id/child_id`.

The two formats disagree, so the `crossplane.io/external-name` annotation flips between them every reconciliation, causing perpetual updates / delete-recreate cycles and AWS `429` throttling.

This adds `vpcEndpointAssociationIdentifier`, which reconstructs both forms deterministically from the named fields:
- **external name**: `vpc_endpoint_id/child_id` (stable import format);
- **Terraform id**: `a-<vpc_endpoint_id><hashcode(child_id)>`, mirroring `terraform-provider-aws` `internal/create.StringHashcode`.

`aws_vpc_endpoint_security_group_association` already uses `config.IdentifierFromProvider` (the `a-<hash>` id on both sides) and is consistent, so it is left unchanged. The now-redundant `uptest.upbound.io/disable-import` annotations on the subnet association examples are removed since import is stable again.

Fixes #2116

### I have:

- [x] Read and followed the [contribution process](https://github.com/crossplane-contrib/provider-upjet-aws/blob/main/CONTRIBUTING.md).
- [x] Run `go build ./config/...` and `go vet ./config/` (both pass).
- [x] Added/updated examples (removed obsolete `disable-import` annotations).
- [ ] Run `make reviewable` to ensure this PR is ready for review (codegen unaffected — external-name config does not change CRDs).

### How has this code been tested?

- Verified against the upstream `terraform-provider-aws` source that the internal id is `a-<vpc_endpoint_id><hashcode(child_id)>` (via `vpcEndpoint*AssociationCreateID`) and the import id is `vpc_endpoint_id/child_id`.
- Reproduced the oscillation live on a managed `VPCEndpointSubnetAssociation` (external-name alternating between the two forms with continuously climbing `resourceVersion`); the sibling `SecurityGroupAssociation` using `IdentifierFromProvider` was stable, confirming the diagnosis.
- `go build` / `go vet` / `gofmt` clean.


aditmeno:
@erhancagirici Could you review my PR?

jonasz-lasut:
/test-examples="examples/ec2/cluster/v1beta1/vpcendpointsubnetassociation.yaml"

jonasz-lasut:
Hi @aditmeno thank you for opening the pull request, I'll try to review it this week

jonasz-lasut:
/test-examples="examples/ec2/cluster/v1beta1/vpcendpointroutetableassociation.yaml"

jonasz-lasut (approved):
LGTM! Thank you @aditmeno for the contribution
(weird, I can't merge it with admin permissions, will go back to this PR and check)

aditmeno:
> * bobh66, erhancagirici, jastang, sergenyalcin, turkenf, and/or ulucinar

It looks like a review from one of these maintainers is needed for the merge

config/externalname.go
```diff
@@ -3270,6 +3271,77 @@ func FormattedIdentifierFromProvider(separator string, keys ...string) config.Ex
...
+	return s, nil
+}
+
+// ec2StringHashcode mirrors terraform-provider-aws internal/create.StringHashcode,
+// which the AWS provider uses to build VPC endpoint association Terraform IDs.
+func ec2StringHashcode(s string) int {
+	v := int(crc32.ChecksumIEEE([]byte(s)))
+	if v >= 0 {
+		return v
+	}
```
bobh66, line 3337: Is this equivalent to:

```suggestion
	} else {
	return -v
	}
```

erhancagirici: AFAIU, this was to check/prevent overflow after negation and handle the case where v == MinInt, which in that case returns 0, so I believe it should stay as is.

jonasz-lasut: That was my understanding as well

aditmeno: It's not quite equivalent — the branch is intentional.

`ec2StringHashcode` is a verbatim port of terraform-provider-aws `internal/create.StringHashcode`, kept line-for-line identical so the reconstructed id `a-<vpc_endpoint_id><hashcode(child_id)>` matches the AWS provider's exactly; any divergence would reintroduce the oscillation this PR fixes.

For `v == math.MinInt`, negation overflows (`-math.MinInt == math.MinInt`, still negative), so `-v >= 0` is false and it falls through to `return 0` — a plain `return -v` would return a negative hashcode there. (On 64-bit `int(crc32.ChecksumIEEE(...))` is always non-negative, so the branch is effectively unreachable, but I'd keep it faithful to upstream.)

erhancagirici (approved):

PR #2122, opened 2026-06-17
Decouple conversion webhook registration from leader election

sergenyalcin:

Applies the upjet template fix for the conversion webhook / leader election bug (upjet#<N>) to this provider.

- **`hack/main.go.tmpl`**: removes `StartWebhooks` from both opts structs; adds an explicit `startWebhooks` block before the `canSafeStart` gate that calls `SetupWebhookWithManager_<group>` for both cluster-scoped and namespaced controllers. A comment at the call site explains why webhook registration must happen here rather than inside the gate.

- **`config/templates/controller.go.tmpl`**: mirrors the upjet template change — adds `SetupWebhookWithManager`, removes the `if o.StartWebhooks { ... }` block from `Setup`.

- **Generated files**: all `zz_main.go`, `zz_controller.go`, and `zz_*_setup.go` files regenerated from the updated templates.

Note: This PR uses the forked upjet dependency until the upjet PR is merged.

I have:

- [ ] Not made any manual changes to generated files, and verified this with `make check-diff`.

https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/27750948935


sergenyalcin:
/test-examples="examples/ec2/cluster/v1beta1/vpc.yaml"

hack/main.go.tmpl
```diff
@@ -271,6 +269,15 @@ func main() {
 		namespacedOptions.ChangeLogOptions = &clo
 	}
 
+	// Webhooks are registered eagerly on all pods before mgr.Start() so that
+	// every replica (leader and followers alike) can serve conversion requests.
+	// Reconciler setup is deferred to the gate and only runs on the leader.
+	startWebhooks := *certsDir != ""
+	if startWebhooks {
```
ulucinar, line 276: nit:

```suggestion
	if *certsDir != "" {
```

ulucinar (commented):
Thank you @sergenyalcin for working debugging & fixing this issue.

hack/main.go.tmpl
sergenyalcin: I've intentionally preferred this version because I think using a variable is better to understand the condition here.

ulucinar (approved):

PR #2125, opened 2026-06-18
Add SourceIdentity option to ProviderConfig assumeRoleChain

rhysxevans:
## Description

Adds an optional `sourceIdentity` field to each entry of
`ProviderConfig.spec.assumeRoleChain`. When set, the value is passed
through to the AWS SDK's `stscreds.AssumeRoleOptions.SourceIdentity`,
which makes it the session-level [`sts:SetSourceIdentity`](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html#API_AssumeRole_RequestParameters)
parameter on the resulting assumed-role session.

Once set, the value appears in CloudTrail as
`userIdentity.sessionContext.sourceIdentity` and can be referenced in
IAM trust policies via the [`aws:SourceIdentity`](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-keys.html#condition-keys-sourceidentity)
condition key, which is useful for end-to-end traceability when many
Crossplane Claims share the same controller pod / role chain.

Relates to #2099 (kept as non-closing so reviewers can decide whether
this fully resolves the issue or whether additional changes — e.g.
Composition-level plumbing — are needed first).

## Changes

| Area | File(s) |
| --- | --- |
| API: cluster-scoped v1beta1 | `apis/cluster/v1beta1/types.go` (+ generated deepcopy) |
| API: namespaced v1beta1 | `apis/namespaced/v1beta1/types.go` (+ generated deepcopy) |
| Client wiring | `internal/clients/provider_config.go` |
| Tests | `internal/clients/provider_config_test.go` (new) |
| CRDs (regenerated) | `package/crds/aws.upbound.io_providerconfigs.yaml`, `package/crds/aws.m.upbound.io_providerconfigs.yaml`, `package/crds/aws.m.upbound.io_clusterproviderconfigs.yaml` |
| Docs | `AUTHENTICATION.md` (usage example + STS permission requirements) |

Total: 10 files, +183 / -0.

## How it was tested

- `go test ./internal/clients/...` — new unit tests cover both the
  positive path (SourceIdentity propagated into `AssumeRoleOptions`)
  and the empty-value path (no propagation).
- Built a custom `linux_arm64` xpkg from this branch, pushed to a
  private ECR, and deployed onto a POC EKS cluster (Crossplane
  v2, Pod Identity). After replacing the public family + sqs packages
  with the patched images:
  - `providers.pkg.crossplane.io` for both family and sqs settled to
    `INSTALLED=True HEALTHY=True`.
  - `kubectl explain providerconfigs.spec.assumeRoleChain.sourceIdentity`
    returned the new field with the expected description.
  - A server-side dry-run apply of a `ProviderConfig` containing
    `sourceIdentity: test-source-identity` was accepted by the API
    server.

Live CloudTrail verification (observing
`userIdentity.sessionContext.sourceIdentity` on an actual reconciled
managed resource) is being tracked separately and is not blocking
this change.

## Caveats

- This is a purely additive schema change; existing `ProviderConfig`
  manifests continue to work unchanged.
- Setting `sourceIdentity` at runtime requires `sts:SetSourceIdentity`
  on both the caller's identity policy and the assumed role's trust
  policy. This is called out in `AUTHENTICATION.md`.
- Upgrading an existing cluster that already has the
  `providerconfigs.aws.upbound.io` CRD installed by an older family
  revision can hit a Crossplane v2 MRD/CRD reconciliation edge-case
  where the CRD spec is not refreshed (`PendingManagedResource`).
  This is a Crossplane core behavior, not specific to this change.

## Checklist

- [x] Code compiles
- [x] Unit tests added
- [x] CRDs regenerated from the API changes
- [x] Docs updated
- [x] DCO sign-off on the commit


rhysxevans:
The composition looks like 

```yaml
# The Composition synthesises a per-claim ProviderConfig so that the
# patched provider-aws (FR crossplane-contrib/provider-upjet-aws#2099)
# can set sts:SourceIdentity from the originating principal carried on
# the XR's spec.parameters.sourceIdentity. The Queue is then bound to
# that per-claim ProviderConfig via spec.providerConfigRef.name.
apiVersion: apiextensions.crossplane.io/v1
kind: Composition
metadata:
  name: xsqsqueues.api.io
  labels:
    crossplane.io/xrd: xsqsqueues.api.io
spec:
  compositeTypeRef:
    apiVersion: api.io/v1alpha1
    kind: XSQSQueue
  mode: Pipeline
  pipeline:
    - step: patch-and-transform
      functionRef:
        name: function-patch-and-transform
      input:
        apiVersion: pt.fn.crossplane.io/v1beta1
        kind: Resources
        resources:
          - name: providerconfig
            base:
              kind: ProviderConfig
              spec:
                credentials:
                  source: PodIdentity
                assumeRoleChain:
                  - roleARN: arn:aws:iam::012345678910:role/foo-poc-crossplane-sqs
            patches:
              - type: FromCompositeFieldPath
                fromFieldPath: metadata.name
                toFieldPath: metadata.name
                transforms:
                  - type: string
                    string:
                      type: Format
                      fmt: "sqs-%s"
              # SourceIdentity is Optional: when the originating principal is
              # absent (legacy claims, system requests) the patch no-ops and
              # the provider omits the field on its sts:AssumeRole call.
              - type: FromCompositeFieldPath
                fromFieldPath: spec.parameters.sourceIdentity
                toFieldPath: spec.assumeRoleChain[0].sourceIdentity
                policy:
                  fromFieldPath: Optional
          - name: queue
            base:
              apiVersion: sqs.aws.upbound.io/v1beta1
              kind: Queue
              spec:
                forProvider: {}
                providerConfigRef:
                  name: default
            patches:
              - type: FromCompositeFieldPath
                fromFieldPath: spec.parameters.region
                toFieldPath: spec.forProvider.region
              - type: FromCompositeFieldPath
                fromFieldPath: spec.parameters.tags
                toFieldPath: spec.forProvider.tags
              # Set spec.forProvider.name so the SQS queue name matches the
              # foo-poc-* IAM resource constraint on the crossplane-sqs role.
              # (The Terraform-based provider uses forProvider.name, not external-name,
              # to control the queue name.)
              - type: FromCompositeFieldPath
                fromFieldPath: metadata.name
                toFieldPath: spec.forProvider.name
                transforms:
                  - type: string
                    string:
                      type: Format
                      fmt: "foo-poc-%s"
              # Bind the Queue to the per-claim ProviderConfig synthesised above.
              - type: FromCompositeFieldPath
                fromFieldPath: metadata.name
                toFieldPath: spec.providerConfigRef.name
                transforms:
                  - type: string
                    string:
                      type: Format
                      fmt: "sqs-%s"
              - type: ToCompositeFieldPath
                fromFieldPath: status.atProvider.url
                toFieldPath: status.atProvider.url
              - type: ToCompositeFieldPath
                fromFieldPath: status.atProvider.id
                toFieldPath: status.atProvider.arn
```

PR #2137, opened 2026-07-02
Fix Stale Resource Tag Keys in status.AtProvider.tags and status.AtProvider.tagsAll

ulucinar:

Relevant PRs: https://github.com/crossplane/upjet/pull/693

This PR fixes an old issue where a resource tag removed from an MR's `spec.forProvider.tags` would keep appearing in `status.atProvider.tags` and `status.atProvider.tagsAll`. 

An example scenario is when you provision a resource with the `spec.forProvider.tags`:
```yaml
...
spec:
  forProvider:
    tags: {"environment": "test-env"}
```
, and later decide to rename the existing tag key as `Environment` so that the intent now becomes:
```yaml
spec:
  forProvider:
    tags: {"Environment": "test-env"}
```

Before this fix, after the external AWS resource acquires the `environment` tag and the resource tags have been observed and recorded in `status.atProvider.tags` and `status.atProvider.tagsAll`, if you replace the existing tag key as `Environment`, the external resource's `environment` tag would successfully be replaced as `Environment` (so no bugs on the external resource itself), but you would observe something like the following on the managed resource:
```yaml
...
spec:
  forProvider:
    tags:
      Environment: test-env
      crossplane-kind: cluster.rds.aws.upbound.io
      crossplane-name: example-alper
      crossplane-providerconfig: default
...
status:
  atProvider:
    tags:
      Environment: test-env
      environment: test-env
      crossplane-kind: cluster.rds.aws.upbound.io
      crossplane-name: example-alper
      crossplane-providerconfig: default
    tagsAll:
      Environment: test-env
      environment: test-env
      crossplane-kind: cluster.rds.aws.upbound.io
      crossplane-name: example-alper
      crossplane-providerconfig: default
```
Please note how the now-removed `environment` tag already appears in status.

The root cause of this issue is the MR's `status.atProvider` map already contains the removed key when the new state is deserialized into it, merging the old and the new keys. For tags, this is not the desired behavior, they should be replaced. This PR achieves it by introducing a custom terraformed template that assigns the zero value for the `Status.AtProvider.Tags` and `Status.AtProvider.TagsAll` fields for MRs that have those fields in `SetObservation` before unmarshaling the new state.

I have:

- [x] Bump upjet once https://github.com/crossplane/upjet/pull/693 is merged.

Manually tested with a `Cluster.rds` resource using https://github.com/crossplane/upjet/pull/693.

### DIFF
--- a/apis/cluster/accessanalyzer/v1beta2/zz_analyzer_terraformed.go
+++ b/apis/cluster/accessanalyzer/v1beta2/zz_analyzer_terraformed.go
@@ -36,6 +36,8 @@ func (tr *Analyzer) GetObservation() (map[string]any, error) {
 
 // SetObservation for this Analyzer
 func (tr *Analyzer) SetObservation(obs map[string]any) error {
+	tr.Status.AtProvider.Tags = nil
+	tr.Status.AtProvider.TagsAll = nil
 	p, err := json.TFParser.Marshal(obs)
 	if err != nil {
 		return err
--- a/apis/cluster/acm/v1beta2/zz_certificate_terraformed.go
+++ b/apis/cluster/acm/v1beta2/zz_certificate_terraformed.go
@@ -36,6 +36,8 @@ func (tr *Certificate) GetObservation() (map[string]any, error) {
 
 // SetObservation for this Certificate
 func (tr *Certificate) SetObservation(obs map[string]any) error {
+	tr.Status.AtProvider.Tags = nil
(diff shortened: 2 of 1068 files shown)


ulucinar:

ulucinar:
/test-examples="examples/rds/cluster/v1beta2/cluster.yaml"

erhancagirici (approved):
Thanks @ulucinar ! LGTM

PR #2140, opened 2026-07-10
Add custom diff and custom late initialize function for elbv2.LBListenerRule resource

sergenyalcin:
### Problem

`aws_lb_listener_rule` resources using the `action` and `action.forward` block trigger a perpetual reconcile loop, every Observe/Update cycle produces a diff even when nothing has changed.

The Terraform provider's `flattenForwardAction*` function attempts to distinguish between different operation contexts (create/update vs. refresh vs. import) by inspecting low-level `GetRawConfig()` and `GetRawPlan()` values. In native Terraform these differ per operation, so the provider uses their null-ness as a signal: null `RawPlan`/`RawConfig` → treat as refresh/import → write both `target_group_arn` and forward to state (`flattenForwardActionBoth`); forward present in `RawConfig` → treat as normal plan/apply → strip `target_group_arn` (`flattenForwardActionOneOf`).

This assumption is incompatible with upjet. Upjet doesn't map to Terraform's operation primitives, all reconciliation flows through the operation calls.

Late-init then copies it to spec. On the following `Observe()`, `RawConfig` is populated (as upjet sets it explicitly), so `flattenForwardActionOneOf` fires and strips `target_group_arn` from state, leaving spec with the arn and state without it, producing a diff that triggers another `Update()`. The cycle repeats indefinitely.

A secondary issue: the generic `LateInitializer` skips the entire action slice once any field is set, so auto-assigned fields like order and provider-defaulted forward.stickiness values are never written back to spec.

### Fix

Forced late-init (`lbListenerRuleForcedLateInitAction`): Re-runs late-init at the individual action-element level, bypassing the "skip non-nil slice" behavior. Also recurses into the Forward sub-block to populate nil sub-fields.

Custom diff (`lbListenerRuleCustomDiff`): Suppresses the remaining spurious Update caused by the RawPlan cycle:

- Case A (Old="", New=arn): suppress when forward is present in the ResourceConfig for that action index — the provider stripped arn via flattenForwardActionOneOf and the arn in spec is a late-init artifact.
- Case B (Old=arn, New="", NewRemoved): suppress when forward is active in the observed state — arn was backfilled by flattenForwardActionBoth, not user-specified.

Real arn changes (actual user edits) are never suppressed.

I have:

- Cluster scoped v1beta1 -> https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/29194425501
- Cluster scoped v1beta2 -> https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/29198145181
- Namespaced v1beta1 -> https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/29199070538

### DIFF
--- a/apis/cluster/elbv2/v1beta1/zz_lblistenerrule_terraformed.go
+++ b/apis/cluster/elbv2/v1beta1/zz_lblistenerrule_terraformed.go
@@ -120,10 +120,58 @@ func (tr *LBListenerRule) LateInitialize(attrs []byte) (bool, error) {
 	opts := []resource.GenericLateInitializerOption{resource.WithZeroValueJSONOmitEmptyFilter(resource.CNameWildcard)}
 
 	li := resource.NewGenericLateInitializer(opts...)
-	return li.LateInitialize(&tr.Spec.ForProvider, params)
+	changed, err := li.LateInitialize(&tr.Spec.ForProvider, params)
+	if err != nil {
+		return false, err
+	}
+	forcedChanged, err := lbListenerRuleForcedLateInitAction(&tr.Spec.ForProvider, params)
+	if err != nil {
+		return false, err
+	}
+	return changed || forcedChanged, nil
 }
 
 // GetTerraformSchemaVersion returns the associated Terraform schema version
 func (tr *LBListenerRule) GetTerraformSchemaVersion() int {
 	return 0
+}
+
+// lbListenerRuleForcedLateInitAction forces late-initialization of nil fields
+// within each element of the action slice. The generic late-initializer skips
+// the entire action slice once it is non-nil (i.e. the user set any action
+// field), so auto-assigned fields like order are never written back to spec,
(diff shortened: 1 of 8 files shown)


sergenyalcin:
/test-examples="examples/elbv2/cluster/v1beta1/lblistenerrule.yaml"

sergenyalcin:
/test-examples="examples/elbv2/cluster/v1beta2/lblistenerrule.yaml"

sergenyalcin:
/test-examples="examples/elbv2/namespaced/v1beta1/lblistenerrule.yaml"

jonasz-lasut (approved):
I've spent the past hour trying to understand and possibly simplify this solution as it's one of the more complex late-initialization/custom diff configurations in upjet providers. I was not able to simplify it further and it's really well documented so from my side it's approved! Thank you @sergenyalcin for all time that went into discovery and implementation here

erhancagirici (approved):

PR #2147, opened 2026-07-14
examples/rds: stop hardcoding engineVersion in Instance examples

ulucinar:

The `rds.aws.upbound.io` `Instance` examples hardcoded a specific `engineVersion`
(e.g. `"16.6"`, `"13.7"`, `"5.7"`). Once RDS retires that exact version, applying
the example fails. `engineVersion` is optional on the `Instance` CRD, so
it can safely be omitted and left for AWS to default.

This PR:

- **Removes the hardcoded `engineVersion`** from all 14 `Instance` examples under
  `examples/rds/` (cluster + namespaced, `v1beta1`/`v1beta2`/`v1beta3`), including
  the ones where the `Instance` is a dependency (`snapshot`, `dbsnapshotcopy`,
  `instancestate`, `instanceroleassociation`, `dbinstanceautomatedbackupsreplication`).
  AWS then selects a currently-available default engine version at creation time.
- In the two `dbinstanceautomatedbackupsreplication.yaml` examples, **also removes
  the coupled `parameterGroupName: default.mysql5.7`.** That default parameter group
  is only valid for a MySQL 5.7 instance, so dropping only `engineVersion` would
  leave an engine-family mismatch. With both removed, AWS attaches the matching
  default parameter group for whichever engine version it selects.
- **Adds two new `v1beta3` (latest cluster-scoped) `Instance` examples** that
  intentionally demonstrate *explicit* version pinning, for testing those configurations,
  and document the coupling that must be kept consistent:
  - `instance-engineversion.yaml`: pins `engineVersion` and pairs it with the
    matching AWS default parameter group (`default.postgres16`).
  - `instance-custom-parametergroup.yaml`: attaches a custom, provider-managed
    `ParameterGroup` via `parameterGroupNameSelector`. There was previously no
    example wiring an `Instance` to an `rds.aws.upbound.io` `ParameterGroup`.
    Because the parameter group has a fixed `family` (`postgres16`), the engine
    version is pinned to match it.

I have:

### DIFF
--- a/examples/rds/cluster/v1beta1/instancestate.yaml
+++ b/examples/rds/cluster/v1beta1/instancestate.yaml
@@ -36,7 +36,6 @@ spec:
     backupWindow: 09:46-10:16
     name: example
     engine: postgres
-    engineVersion: "16.6"
     instanceClass: db.t3.micro
     kmsKeyIdSelector:
       matchLabels:
--- a/examples/rds/cluster/v1beta1/instancestate.yaml
+++ b/examples/rds/cluster/v1beta1/instancestate.yaml
@@ -34,7 +34,7 @@ spec:
     autoMinorVersionUpgrade: true
     backupRetentionPeriod: 14
     backupWindow: 09:46-10:16
-    name: example
+    dbName: example
     engine: postgres
     instanceClass: db.t3.micro
     kmsKeyIdSelector:
--- a/examples/rds/namespaced/v1beta1/instancestate.yaml
+++ b/examples/rds/namespaced/v1beta1/instancestate.yaml
@@ -38,7 +38,6 @@ spec:
     backupWindow: 09:46-10:16
     name: example
     engine: postgres
-    engineVersion: "16.6"
     instanceClass: db.t3.micro
     kmsKeyIdSelector:
       matchLabels:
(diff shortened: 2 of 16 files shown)


ulucinar:
/test-examples="examples/rds/cluster/v1beta3/instance.yaml"

erhancagirici (dismissed):

examples/rds/cluster/v1beta1/instancestate.yaml
```diff
@@ -36,7 +36,6 @@ spec:
     backupWindow: 09:46-10:16
     name: example
```
erhancagirici: nit: not introduced in this PR, but raising for the sake of completeness
```suggestion
    dbName: example
```

examples/rds/namespaced/v1beta1/instancestate.yaml
```diff
@@ -38,7 +38,6 @@ spec:
     backupWindow: 09:46-10:16
     name: example
```
erhancagirici: nit: not introduced in this PR, but raising for the sake of completeness
```suggestion
    dbName: example
```

erhancagirici (approved):

PR #2164, opened 2026-07-28
Add `RuntimeManagementConfig.lambda` resource

### LINKED ISSUE #2157: Request for `aws_lambda_runtime_management_config` resource

Terraform Resource Name: `aws_lambda_runtime_management_config`

This resource manages a Lambda function's [runtime management configuration](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-update.html) (`PutRuntimeManagementConfig`): whether the managed runtime updates automatically (`Auto`), on function update (`FunctionUpdate`), or is pinned to a specific runtime version ARN (`Manual`). It is present in the v6.34.0+ Terraform AWS provider that `provider-upjet-aws` currently builds against, but no external-name configuration exists for it, so no CRD/MRD is generated in the `lambda` group.

### What is your use case?

We manage AWS Lambda functions declaratively with the `lambda` provider family. A recent AWS refresh of a managed runtime introduced a behavioral regression for a subset of our functions, and the mitigation AWS documents for this situation is runtime management controls: pin `update_runtime_on = "Manual"` with a known-good `runtime_version_arn` until affected functions are re-bundled, then return them to `Auto`.

Because this provider does not expose the runtime management configuration, our operator currently calls the Lambda API directly (`GetRuntimeManagementConfig` / `PutRuntimeManagementConfig`) alongside the Crossplane-managed `Function` resources. That splits one function's configuration across two reconciliation mechanisms with different drift behavior. Generating `aws_lambda_runtime_management_config` would let the runtime pin live next to the `Function` MR it belongs to.

We're happy to contribute a PR adding it.

sac-outsystems:

Adds the `RuntimeManagementConfig.lambda` managed resource (`aws_lambda_runtime_management_config`), which pins how a Lambda function adopts new managed runtime versions (`Auto`, `FunctionUpdate`, or `Manual`).

This is a terraform-plugin-framework resource with no `id` attribute; it is imported as `function_name,qualifier`. The external-name is therefore computed from those fields via a small `lambdaRuntimeManagementConfig()` helper (`qualifier` defaults to `$LATEST`). `function_name` (with a reference to `aws_lambda_function`) and `qualifier` remain regular spec fields so the reference resolves and an explicit qualifier reaches Terraform.

Fixes #2157

I have:

- [x] Run `make generate` and committed the results (in a separate commit).

`make reviewable` passes (generate is drift-free, lint and unit tests green).

Manually tested end-to-end against a real AWS account with the locally built provider (the Lambda group has no automated uptest, since a Function needs a local zip):

- **Create** with `updateRuntimeOn: FunctionUpdate` -> MR reaches `Ready/Synced=True`; `aws lambda get-runtime-management-config` reports `FunctionUpdate`.
- **Update** `FunctionUpdate` -> `Auto` (and back) -> reconciles cleanly; AWS reflects each change.
- **Manual pinning** with `updateRuntimeOn: Manual` and a real `runtimeVersionArn` (read from the function's `INIT_START` CloudWatch log line) -> MR reaches `Ready/Synced=True`; `aws lambda get-runtime-management-config` reports `Manual` with the pinned `RuntimeVersionArn`.
- **Delete** -> AWS resets the function to `Auto`. The TF resource embeds `framework.WithNoOpDelete` but overrides it with an explicit `Delete` that calls `PutRuntimeManagementConfig(Auto)`, so removing the resource resets the managed-runtime mode to `Auto`.


jonasz-lasut:
Hi @sac-outsystems thank you for opening the PR, could you update the example to be uptestable without manual intervention annotation similarly as `examples/lambda/cluster/v1beta2/function.yaml` deploys lambda? We have a predefined test s3 bucket with example application for e2e purpose called `official-provider-test-data`. Really appreciate the contribution!

sac-outsystems:
Hi @jonasz-lasut, thanks for the review! 
I've updated both examples (cluster and namespaced) to be self-contained and uptestable without the manual-intervention, following the function.yaml example you shared.

Each updated example now deploys:
- an IAM Role (Lambda trust policy),
- a Function sourced from the official-provider-test-data bucket (s3Key: hello-python.zip, runtime: python3.9), and
- the RuntimeManagementConfig, which now references the Function via functionNameSelector instead of a hardcoded name.

Let me know if I missed anything.

jonasz-lasut:
/test-examples="examples/lambda/cluster/v1beta1/runtimemanagementconfig.yaml"

jonasz-lasut (approved):
Thank you for your contribution @sac-outsystems ! LGTM

PR #2172, opened 2026-07-29
account: add PrimaryContact managed resource

davemac30:
### Description of your changes

Adds the `PrimaryContact` managed resource to the `account` family, corresponding to the Terraform [`aws_account_primary_contact`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/account_primary_contact) resource.

`aws_account_primary_contact` is a singleton resource — there is exactly one primary contact per AWS account — so it uses `config.IdentifierFromProvider` and the AWS account ID as the external name.

**Usage notes included in the example manifest:**
- For `managementPolicies: [Observe]` or importing an existing resource, `crossplane.io/external-name` must be set to the AWS account ID before applying.
- For normal Create/Update/Delete usage the annotation is populated automatically after the first reconcile.

Fixes #2165

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make generate` and committed the results (ideally in a separate commit). 
- [x] Not made any manual changes to generated files, and verified this with `make check-diff`.

### How has this code been tested

Tested manually against a live AWS account using a local provider binary connected to a `kind` cluster:

- **Observe-only** (`managementPolicies: ["Observe"]`): resource reached `SYNCED=True / READY=True` and `status.atProvider` was populated with the real account contact information.
- **Full CRUD** (`managementPolicies: ["*"]`): applied forProvider fields were written to AWS and reflected back in `status.atProvider`; setting `websiteUrl: ""` correctly cleared the field in AWS.

[contribution process]: https://git.io/fj2m9

PR #2181, opened 2026-08-03
account: add PrimaryContact managed resource

### LINKED ISSUE #2165: Add PrimaryContact resource to provider-aws-account
### Summary

The `provider-aws-account` family provider currently exposes `AlternateContact` and `Region` managed resources, but does not include a `PrimaryContact` resource corresponding to the Terraform `aws_account_primary_contact` resource.

### Use Case

We manage AWS accounts declaratively via Crossplane. Each account requires a primary contact to be set (full name, company, address, phone, etc.) as part of our account bootstrap process. Today we can manage all three **alternate** contacts (billing, operations, security) via `AlternateContact`, but the **primary** contact must be set out-of-band because there is no corresponding Crossplane resource.

### Terraform Resource

The upstream Terraform AWS provider has supported this since v4.27.0:

- [`aws_account_primary_contact`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/account_primary_contact)

It wraps the AWS Account API operations:
- `account:PutContactInformation`
- `account:GetContactInformation`

### Expected Crossplane Resource

```yaml
apiVersion: account.aws.upbound.io/v1beta1
kind: PrimaryContact
metadata:
  name: example
spec:
  forProvider:
    accountId: "123456789012"  # optional, for member accounts
    fullName: "Jane Doe"
    companyName: "Example Corp"
    addressLine1: "123 Main St"
    city: "Seattle"
    stateOrRegion: "WA"
    countryCode: "US"
    postalCode: "98101"
    phoneNumber: "+15555550100"
    websiteUrl: "https://example.com"  # optional
```

### Impact

Without this resource, organisations that bootstrap accounts via Crossplane have a gap in their contact management — alternate contacts are fully declarative but the primary contact requires manual intervention or a separate automation path.

### Additional Context

- The Terraform resource has been stable and widely used since 2022.
- The AWS API (`PutContactInformation` / `GetContactInformation`) supports both standalone accounts and member accounts in an organisation (via the `AccountId` parameter).
- This appears to be a straightforward upjet mapping — the Terraform resource is simple with no complex references or dependencies.

davemac30:

Adds the `PrimaryContact` managed resource to the `account` family, corresponding to the Terraform [`aws_account_primary_contact`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/account_primary_contact) resource.

This is a rework of #2172 (reverted in #2180). The generated files have been regenerated from scratch on top of the current `main` (crossplane-runtime ≥2.3) to eliminate the stale generated output that caused the revert.

`aws_account_primary_contact` is a singleton resource — there is exactly one primary contact per AWS account — so it uses `config.IdentifierFromProvider`.

**External name behaviour (documented in the example manifests):**
- When `forProvider.accountId` is omitted (managing the account the provider credentials belong to), the Terraform provider uses `"default"` as the resource ID, so the external name will be `"default"`.
- When `forProvider.accountId` is set explicitly, the external name is the target AWS account ID. Cross-account management requires the caller's IAM identity to have `account:GetContactInformation` and `account:PutContactInformation` permissions scoped to the target account.
- For `managementPolicies: [Observe]` or import, set `crossplane.io/external-name` to `"default"` (current-account) or the explicit account ID, matching how the resource was originally created.

Fixes #2165

I have:

Tested manually against live AWS accounts using a local provider binary connected to a `kind` cluster.

Two scenarios were tested:

**Scenario A — current-account management (member account managing itself, no `accountId`)**

Apply, UpToDate, Import and Delete all confirmed working. External name `"default"` is correctly assigned.

**Scenario B — cross-account management (management account targeting member account via `accountId`)**

`forProvider.accountId` and `crossplane.io/external-name` both set to the member account ID. `status.atProvider` correctly reflects the member account's contact data (not the management account's), confirming `account_id` is correctly propagated to the AWS API.

Full test report was posted as a comment on #2172.


jonasz-lasut (approved):
Thank you @davemac30 LGTM!

Issue #1444, opened 2024-08-05
[Bug]:aws.rds. provider recouncile loop triggered by setting default parameter explictly


momoXD007:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Affected Resource(s)

rds.aws.upbound.io/v1beta1 - ParameterGroup

### Steps to Reproduce

Install provider, provision a Postgres database, with default parameter explicitly set

### What happened?

Hello,
we are using Crossplane to provison sql and postgres databases and we believe we found a bug. An example claim we had looked like this:
```
[...]
spec:
  [...]
  resourceConfig:
    allowMajorVersionUpgrade: true
    applyImmediately: true
    autoMinorVersionUpgrade: true
    autogeneratePassword: true
    dbName: dbNameHere
    deleteAutomatedBackups: true
    deletionPolicy: Delete
    engine: postgres
    engineVersion: '15'
    iamAuthentication: false
    instanceClass: db.t4g.small
    iops: 3000
    kmsKeyId: >-
      arn:aws:kms:eu-central-1:<someaccout_id>:key/<some key>
    maintenanceWindow: 'Mon:00:00-Mon:03:30'
    maxAllocatedStorage: 0
    multiAz: false
    parameterFamilyGroup: postgres15
    parameterGroup:
      - applyMethod: immediate
        name: log_checkpoints
        value: '1'
    passwordSecretRef:
      key: password
      name: semantic-recommender-rds-secret
      namespace: default
    performanceInsightsEnabled: false
    skipFinalSnapshot: true
    storageType: gp3
    tags:
      Prometheus-Export-Enabled: 'Yes'
      team: exampleTeam
```
We noted that Crossplane was constantly reconciling the DB ParameterGroup for this claim. Crossplane tried to change the parameter “log_checkpoints”. Manually checking this parameter in aws console confirmed that the parameter is by default ‘1’ as this is the engine default.
So in the aws_console the UI shows it like this:
```
Name: | Value: | Source:
log_checkpoints | 1 | Engine default
```
so updating this parameter is useless as it already has the right value by default still Crossplane tries non-stop.
As an experiment we changed it 0 in our claim
This resulted in the following:
```
Name: | Value: | Source:
log_checkpoints | 0 | modified
```
the reouncilitation loop finished.
Now the interesting part we change it back to 1 in our claim:
Crossplane recounciles and we get the following in aws-console:
```
Name: | Value: | Source:
[...]

PR #2182, opened 2026-08-04
fix: suppress ApplyMethod-only diffs for DB parameter groups

### LINKED ISSUE #1286: [Bug]: RDS Parameter Group updates parameter in loop

- [X] I have searched the existing issues

### Affected Resource(s)

ParameterGroup.rds.aws.upbound.io

ParameterGroup.rds.aws.upbound.io

### Steps to Reproduce

Create a new ParameterGroup for mysql 8.0 and set the property "innodb_file_per_table": 1

```yaml
apiVersion: rds.aws.upbound.io/v1beta1
kind: ParameterGroup
metadata:
  name: example-parametergroup
spec:
  forProvider:
    description: example
    family: mysql8.0
    parameter:
      - applyMethod: immediate
        name: innodb_file_per_table
        value: "1"
    region: eu-central-1
```

Crossplane will start updating AWS in loop, this increases a lot the Cloudtrail costs.

image

### What happened?

When a new RDS parameter group is created with property "innodb_file_per_table": 1, crossplane tries to apply this config in loop increasing a lot the cloudtrail costs.

image

I have used a few properties and this is the only one that is returning this problem, however I can not guarantee this is the only property that causes problems.

When I look at cloudtrail:

image

The request shows "isModifiable": false, however this parameter shows as modifiable in the parameter group table:

image
 
So I am not sure this is the issue.

```shell
No errors seen in Crossplane.
```

### Crossplane Version

12.2.2

### Provider Version

0.46

### Kubernetes Version

1.27

### Kubernetes Distribution

EKS

### Additional Info

_No response_

erhancagirici:

Fixes perpetual diff with `ClusterParameterGroup` and `ParameterGroup` resources in `rds` , `docdb` and `neptune` API groups when the parameter values matches the AWS default, but only the `ApplyMethod` differs.

### Root cause

It comes from how AWS treats the `apply_method` of a parameter:

1. For a parameter it has not been asked to modify, AWS reports
   `ApplyMethod: pending-reboot`.
2. AWS refuses to change a parameter's apply method *unless its value also changes*, however the request is accepted and silently ignored, so the `apply_method` does not properly roundtrip after updating.

Also see upstream TF docs on RDS [Problematic Plan Changes](https://github.com/hashicorp/terraform-provider-aws/blob/v6.55.0/website/docs/r/db_parameter_group.html.markdown#problematic-plan-changes) section.

### Fix

Added a custom diff such that it inspects the `parameter` set diff and drops every `parameter.*` attribute **only** when all the parameters diffs are `ApplyMethod` only.

Non-update diffs (empty diff, resource creation, destroy diffs), parameter additions/deletions diffs, parameter value change diffs are kept as is.

Fixes #1286, #1444

I have:

- New manifests added for reproducers
- Validated that update loop occurs before this PR with these new example manifests

Uptest: 

[Uptest-f5937a58a7096309113ba1e771525c2b](https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/30925394534)

```
/test-examples="examples/docdb/cluster/v1beta1/clusterparametergroup-loopfix.yaml,
examples/neptune/cluster/v1beta1/cluster-parameter-group-loopfix.yaml,
examples/neptune/cluster/v1beta1/parameter-group-loopfix.yaml,
examples/rds/cluster/v1beta1/clusterparametergroup-loopfix.yaml,
examples/rds/cluster/v1beta1/parametergroup-loopfix.yaml"
```
After review addressing:
https://github.com/crossplane-contrib/provider-upjet-aws/actions/runs/30998555722 -- Passed

### DIFF
--- a/config/diffutils/customdiff.go
+++ b/config/diffutils/customdiff.go
@@ -1,0 +1,104 @@
+// SPDX-FileCopyrightText: 2026 The Crossplane Authors <https://crossplane.io>
+//
+// SPDX-License-Identifier: CC0-1.0
+
+package diffutils
+
+import (
+	"strings"
+
+	"github.com/crossplane/crossplane-runtime/v2/pkg/errors"
+	"github.com/crossplane/upjet/v2/pkg/config"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+type awsDBParameter struct {
+	Value       string
+	ApplyMethod string
+	Hash        string
+}
+
+func flattenDBParameterSet(tfList []any) map[string]awsDBParameter {
+	dbParams := make(map[string]awsDBParameter, len(tfList))
+	for _, tfMapRaw := range tfList {
+		tfMap, ok := tfMapRaw.(map[string]any)
+		if !ok {
+			continue
+		}
+		name, ok := tfMap["name"].(string)
+		if !ok || name == "" {
+			continue
+		}
+		dbParam := awsDBParameter{}
+		// Parameter values are case-sensitive, hence stored verbatim.
+		if v, ok := tfMap["value"].(string); ok {
+			dbParam.Value = v
+		}
+		if v, ok := tfMap["apply_method"].(string); ok && v != "" {
+			dbParam.ApplyMethod = strings.ToLower(v)
+		}
+		dbParams[name] = dbParam
+	}
+
+	return dbParams
+}
+
+// SuppressAWSParameterGroupDiff suppresses diff for AWS DB parameter group diff when the diff is
+// only on `ApplyMethod` causing perpetual diff.
+// See https://github.com/hashicorp/terraform-provider-aws/blob/v6.55.0/website/docs/r/db_parameter_group.html.markdown#problematic-plan-changes
+func SuppressAWSParameterGroupDiff(r *config.Resource, diff *terraform.InstanceDiff, s *terraform.InstanceState) (*terraform.InstanceDiff, error) { //nolint:gocyclo // easier to follow as a unit
+	if isNonUpdateDiff(diff, s) {
+		return diff, nil
+	}
+	d, err := schema.InternalMap(r.TerraformResource.Schema).Data(s, diff)
+	if err != nil {
+		return nil, errors.Wrapf(err, "cannot construct diff data")
+	}
+
+	if d.HasChange("parameter") {
+		o, n := d.GetChange("parameter")
+		os, ns := o.(*schema.Set), n.(*schema.Set)
+
+		// no diff suppression when parameters are added or removed
+		if os.Len() != ns.Len() {
+			return diff, nil
+		}
+
+		oldParams := flattenDBParameterSet(os.List())
+		newParams := flattenDBParameterSet(ns.List())
+
+		// check if flattening changes the parameter count, meaning
+		// there are duplicate parameters with same name and different
+		// values/apply-methods. This can be considered as an invalid
+		// configuration.
+		if len(oldParams) != os.Len() || len(newParams) != ns.Len() {
+			// no diff suppression, let other layers reject if needed
+			return diff, nil
+		}
+		for name, newVal := range newParams {
+			oldVal, ok := oldParams[name]
+			if !ok || oldVal.Value != newVal.Value {
+				// Desired config introduced a new parameter,
+				// or value changed for an existing parameter
+				return diff, nil
+			}
+		}
+
+		// At this point, we validated that the diff for all parameters
+		// in `parameter` attribute is apply-method only.
+		// Suppress this diff.
+		for k := range diff.Attributes {
+			if strings.HasPrefix(k, "parameter.") {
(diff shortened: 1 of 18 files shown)


erhancagirici:
/test-examples="examples/docdb/cluster/v1beta1/clusterparametergroup-loopfix.yaml,examples/neptune/cluster/v1beta1/cluster-parameter-group-loopfix.yaml,examples/neptune/cluster/v1beta1/parameter-group-loopfix.yaml,examples/rds/cluster/v1beta1/clusterparametergroup-loopfix.yaml,examples/rds/cluster/v1beta1/parametergroup-loopfix.yaml"

erhancagirici:
/test-examples="examples/docdb/cluster/v1beta1/clusterparametergroup-loopfix.yaml,examples/neptune/cluster/v1beta1/cluster-parameter-group-loopfix.yaml,examples/neptune/cluster/v1beta1/parameter-group-loopfix.yaml,examples/rds/cluster/v1beta1/clusterparametergroup-loopfix.yaml,examples/rds/cluster/v1beta1/parametergroup-loopfix.yaml"

config/diffutils/customdiff.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "cannot construct diff data")
+	}
+
+	if d.HasChange("parameter") {
+		o, n := d.GetChange("parameter")
+		os, ns := o.(*schema.Set), n.(*schema.Set)
+
+		// no diff suppression when parameters are added or removed
+		if os.Len() != ns.Len() {
```
jonasz-lasut, line 64: **nit:** `flattenDBParameterSet` keys both maps by parameter name, so duplicate-name spec entries *collapse* into a single map entry. When that collapse coincides with a parameter removal in the same edit, this length guard still passes (`apply_method` is part of the set hash, so the duplicates are **distinct** set elements), the new→old loop below finds every surviving name value-unchanged, and the whole `parameter.*` diff — including the removal — is suppressed.

tested it locally using this PR's own test helpers — state `[innodb_file_per_table=1/immediate, max_connections=150/immediate]`, desired config `[innodb_file_per_table=1/immediate, innodb_file_per_table=1/pending-reboot]`:

```
diff before: 9 parameter.* attributes
  parameter.2882656119.name: old="max_connections" new="" removed=true
  parameter.2882656119.value: old="150" new="" removed=true
  parameter.2882656119.apply_method: old="immediate" new="" removed=true
  parameter.2584704414.name: old="" new="innodb_file_per_table" removed=false
  ...
diff after: 0 parameter.* attributes remain
```

The generated CRDs' `parameter` field is an atomic list with no uniqueness constraint, so the API server accepts the duplicate — reachable via e.g. a kustomize/helm list-merge that **appends** an overlay entry instead of *replacing* it.

<details>
<summary>MR reproducer (two steps)</summary>

Step 1 — applied and reconciled:

```yaml
apiVersion: rds.aws.upbound.io/v1beta1
kind: ParameterGroup
metadata:
  name: dup-example
spec:
  forProvider:
    region: us-west-2
    family: mysql8.0
    description: duplicate-name false-suppression reproducer
    parameter:
      - name: innodb_file_per_table
        value: "1"
        applyMethod: immediate
      - name: max_connections
        value: "150"
        applyMethod: immediate
```

Step 2 — spec edited to:

```yaml
    parameter:
      - name: innodb_file_per_table
        value: "1"
        applyMethod: immediate
      - name: innodb_file_per_table
        value: "1"
        applyMethod: pending-reboot
```

</details>

"AWS rejects duplicate parameter names" doesn't cover this path: suppression means `ModifyDBParameterGroup` is **never called**, so the rejection never happens — the MR reports `Synced=True` while `max_connections=150` stays set remotely, and every reconcile repeats the suppression until some other change produces a kept diff. All other duplicate-name shapes (any value change, or a length mismatch) keep the diff and surface a visible AWS error, which is why I'd grade this a hardening nit rather than a realistic-usage correctness bug: it needs a self-contradictory spec *plus* a simultaneous removal in the same edit.

A one-line guard after the two `flattenDBParameterSet` calls closes it:

```go
// Duplicate names collapse when keying by name; don't suppress what we can't correlate.
if len(oldParams) != os.Len() || len(newParams) != ns.Len() {
    return diff, nil
}
```

Happy to leave as-is if you'd rather not carry the guard — non-blocking either way. If you take it, the scenario above drops in as a table case (`want: suppressed: false`) using the existing `param`/`paramGroupState`/`paramGroupDiff` helpers.

jonasz-lasut (approved):
Thank you @erhancagirici LGTM! Left one small nit for you to consider

erhancagirici:
/test-examples="examples/docdb/cluster/v1beta1/clusterparametergroup-loopfix.yaml,examples/neptune/cluster/v1beta1/cluster-parameter-group-loopfix.yaml,examples/neptune/cluster/v1beta1/parameter-group-loopfix.yaml,examples/rds/cluster/v1beta1/clusterparametergroup-loopfix.yaml,examples/rds/cluster/v1beta1/parametergroup-loopfix.yaml"

config/diffutils/customdiff.go
erhancagirici: done 👍

PR #2183, opened 2026-08-07
fix: delete empty trailing YAML documents from examples

barunavo:
## Problem

`examples/dynamodb/cluster/v1beta2/table-with-encryption.yaml` ends with a bare `---` and no content after it, opening a third YAML document with nothing in it.

`yaml.safe_load_all()` parses that as `None`. Lenient consumers skip an empty trailing document silently, which is likely why this went unnoticed. `crank`'s parser does not.

## Fix

1. Removed the trailing blank line and `---` separator from `table-with-encryption.yaml`.
2. Hardened `scripts/check-examples.py` so this class of bug fails CI instead of merging silently. `load_gvks()` iterated over parsed YAML documents and skipped `None` values without comment. It now collects any file that produced a `None` document, prints the filenames, and exits non-zero.

## Testing

- `crank xpkg build` against a minimal package containing only this example file:
  - Before the fix: `error converting YAML to JSON: yaml: line 1: did not find expected node content`, exit 1.
  - After the fix: builds successfully, exit 0.
- `python3 scripts/check-examples.py <crd-dir> <examples-dir>`: exits 2 and prints the filename against the pre-fix example, exits 0 after the fix.


jonasz-lasut (approved):
Thank you @barunavo LGTM!

PR #2186, opened 2026-08-07
fix(sns): derive platform application import IDs

### LINKED ISSUE #2184: [Bug]: SNS PlatformApplication import ID hard-codes GCM

- [x] I have searched the existing issues

### Affected Resource(s)

* sns.aws.m.upbound.io/v1beta1 - PlatformApplication

```yaml
apiVersion: sns.aws.m.upbound.io/v1beta1
kind: PlatformApplication
metadata:
  name: example-ios-application
  annotations:
    # SNS application name, not ARN.
    crossplane.io/external-name: example-ios-application
spec:
  forProvider:
    region: eu-west-1
    platform: APNS_SANDBOX
    platformCredentialSecretRef:
      name: example-ios-credentials
      key: credential
    platformPrincipalSecretRef:
      name: example-ios-credentials
      key: principal
```
### Steps to Reproduce

1. Create an SNS PlatformApplication with platform: APNS or platform: APNS_SANDBOX.
2. Configure the Crossplane MR external name as the SNS application name.
3. Cause the provider to import the existing resource, for example by losing its local Terraform workspace during provider recovery.
4. Observe the Terraform import ID generated from the external-name configuration.

### What happened?

aws_sns_platform_application is configured in config/externalname.go as:
```go
config.TemplatedStringAsIdentifier(
    "name",
    fullARNTemplate("sns", "app/GCM/{{ .external_name }}"),
)
```
This generates a GCM ARN for every PlatformApplication import, regardless of spec.forProvider.platform.

For an APNS_SANDBOX application named example-ios-application, the provider generates:
```
arn:aws:sns:<region>:<account-id>:app/GCM/example-ios-application
```

The required import ID is:
```
arn:aws:sns:<region>:<account-id>:app/APNS_SANDBOX/example-ios-application
```
Terraform imports SNS PlatformApplications using their ARN.

```shell

```

### Crossplane Version

2.3.4

### Provider Version

2.6.0

### Kubernetes Version

1.35.6

### Kubernetes Distribution

EKS

### Additional Info

Terraform documents ARN-based imports for SNS PlatformApplications:

https://github.com/hashicorp/terraform-provider-aws/blob/v5.100.0/website/docs/r/sns_platform_application.html.markdown#L80-L95

AWS documents GCM, APNS, and APNS_SANDBOX as valid platform values:

https://docs.aws.amazon.com/sns/latest/api/API_CreatePlatformApplication.html

The import template should use the configured platform:
```go
config.TemplatedStringAsIdentifier(
    "name",
    fullARNTemplate("sns", "app/{{ .parameters.platform }}/{{ .external_name }}"),
)
```
A previously proposed implementation exists in closed PR [#1226](https://github.com/crossplane-contrib/provider-upjet-aws/pull/1226)

Separately, recovery after provider workspace state loss depends on Upjet selecting import rather than refresh/create for previously created async resources [#562](https://github.com/crossplane/upjet/pull/562)

alexdor:

Fixes #2184.

Construct the `aws_sns_platform_application` Terraform import ID with `spec.forProvider.platform` instead of hard-coding `GCM`. This lets PlatformApplications rebuild their existing ARN after provider state loss for every platform accepted by AWS.

The external-name template renders the runtime platform into the ARN. Although Upjet normally infers template parameters as identifier fields, this resource clears `ExternalName.IdentifierFields`: platform participates in the import ID but is not the resource's external identifier. This preserves the existing `platform` API schema, `initProvider` field, and Kubernetes validation. Import ID construction returns a clear error when platform is missing, empty, or not a string.

I have:

- [x] Kept generated API artifacts unchanged.

- Added table-driven unit coverage for ADM, APNS, APNS_SANDBOX, GCM, and WNS.
- Added coverage for missing, empty, and non-string platform values.
- Asserted that `platform` is not promoted to an Upjet identifier field.
- Ran `go test ./config`.
- Ran `git diff --check`.
- Attempted `make generate`. Terraform initialization succeeded, but the downloaded Darwin arm64 AWS provider binary failed to start during schema generation. The failed command left no generated-file changes.

### DIFF
--- a/config/externalname.go
+++ b/config/externalname.go
@@ -2689,7 +2689,7 @@ var TerraformPluginSDKExternalNameConfigs = map[string]config.ExternalName{
 	//
 	// SNS platform applications can be imported using the ARN:
 	// arn:aws:sns:us-west-2:0123456789012:app/GCM/gcm_application
-	"aws_sns_platform_application": config.TemplatedStringAsIdentifier("name", fullARNTemplate("sns", "app/GCM/{{ .external_name }}")),
+	"aws_sns_platform_application": snsPlatformApplicationExternalName(),
 	// no import documentation is provided
 	// TODO: we will need to check if normalization is possible
 	"aws_sns_sms_preferences": config.IdentifierFromProvider,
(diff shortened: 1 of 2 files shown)


alexdor:
> Thank you for your contribution, this is the right shape: `platform` becomes part of the import id, `IdentifierFields` stays empty so the CRD does not change, and the external name is still the application name, so existing MRs need no migration. Two changes before I approve:
> 
>     1. Drop the `supportedSNSPlatforms` allowlist and pass `platform` through. The Terraform schema only requires `platform` to be a non-empty string (`internal/service/sns/platform_application.go`: `Required: true, ForceNew: true`, no validation), and SNS accepts more values than the four listed (Baidu, WNS and MPNS are in the SNS developer guide next to ADM, APNS, APNS_SANDBOX and GCM). With the allowlist, an application AWS created with one of those platforms fails in `GetIDFn` the first time the provider rebuilds its state from the annotation (after a restart or a cache miss), which is the same failure mode this PR fixes for non-GCM applications. A wrong value is rejected by `CreatePlatformApplication` anyway, so the list buys nothing. The `UnsupportedPlatform` test case then becomes a positive case for `WNS`.
> 
>     2. Do not default to `GCM` when `platform` is missing; return an error instead. `GetIDFn` receives the merged `forProvider`/`initProvider` parameters, so the fallback only triggers for a manifest that sets `platform` nowhere, which Terraform rejects on create anyway. In the one place it can matter, state reconstruction from the annotation, a silently built `app/GCM/<name>` ARN is exactly the bug that got us here; an explicit `platform is required to build the SNS platform application import id` error is easier to diagnose.
> 
> 
> With those two in I am happy to approve; I'll run the CI and E2E tests after the changes are in place

Thanks for looking into it! The fixes should be in place now 😊

PR #2190, opened 2026-08-12
fix(sfn): correct aws_sfn_activity ARN identifier format

e-desouza:
## What problem does this PR solve?

`aws_sfn_activity`'s external-name template built its ARN with a `/`
before the activity name: `activity/{{ .external_name }}`. AWS Step
Functions activity ARNs use a colon separator instead:
`arn:aws:states:<region>:<account-id>:activity:<name>`. This matches
the pattern already used a few lines below for
`aws_sfn_state_machine` (`stateMachine:{{ .external_name }}`).

With the slash, generated/imported external names for existing
`aws_sfn_activity` resources would not match the real resource ARN.

## How has this code been tested?

- Compared against the state-machine sibling entry in the same file,
  which already uses the colon-separated form.
- Verified against the documented AWS ARN format for Step Functions
  activities.

### DIFF
--- a/config/externalname.go
+++ b/config/externalname.go
@@ -2671,7 +2671,7 @@ var TerraformPluginSDKExternalNameConfigs = map[string]config.ExternalName{
 
 	// sfn
 	//
-	"aws_sfn_activity": config.TemplatedStringAsIdentifier("name", fullARNTemplate("states", "activity/{{ .external_name }}")),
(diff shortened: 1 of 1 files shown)


jonasz-lasut (approved):
Thank you for your contribution @e-desouza LGTM!
Tested locally and all v2 versions have this bug, thank you for resolving it!

PR #2212, opened 2026-08-26
feat: extend credentials cache for AssumeRole configurations

erhancagirici:

This PR extends the credential cache to cover the credential sources:
- Secret + AssumeRoleChain
- WebIdentity 
- WebIdentity + AssumeRoleChain
and refactors cache key derivation so that credential material living
outside the ProviderConfig spec is detected when it rotates.

Before this PR the credential cache in `internal/clients/creds_cache.go` only served
`source: IRSA`. ( prior work at https://github.com/crossplane-contrib/provider-upjet-aws/pull/1235 )

### Cache coverage

`cacheableSource` (`internal/clients/creds_cache.go`)  explicitly gates which sources are :

| Source | Cached | Notes |
| --- | --- | --- |
| `IRSA` | yes | as before |
| `WebIdentity` | yes | retrieval calls `sts:AssumeRoleWithWebIdentity` |
| `Secret` / `Fs` / `Environment` (static creds) | only with `spec.assumeRoleChain` | static credentials cost nothing to retrieve; the chain does |
| `PodIdentity` | no | out of the scope of this PR |
| `Upbound` | no | out of the scope of this PR |

### Cache key refactoring

- Each constructed `aws.Config` , now accompanied by a (optional) `credsFingerprint`. This captures out-of-band data specific to the particular auth source, that is not directly a part of the `ProviderConfig` spec. i.e. credential materials that can change without changes in provider config manifest. Like  Secret's contents, a projected token, the
IRSA environment is keyed by its fingerprint.
- This fingerprint becomes a part of the cache key, alongside with the existing ProviderConfig UID, generation, region and source. 
- The cache key is intentionally "pessimistic". Any spec change in PC spec or fingerprint will result in a fresh credential retrieval & new cache entry.
- If a source has out-of-spec material but no fingerprint could be computed, the cache is
skipped rather than risking credentials derived from stale material.

### WebIdentity token reads 

`WebIdentity` with `spec.credentials.webIdentity.tokenConfig` reads its token from a Secret
or the filesystem, which needs care once the credential provider outlives the reconcile
that built it. 

AWS SDK's `TokenRetriever` interface has no `ctx` arg . In the XP provider impl, the WebIdentity token source can be a k8s Secret, which would be appropriate to have context'ed read.

Before this PR (pre-caching), to overcome this limitation, the custom token retriever (`xpWebIdentityTokenRetriever`) implementation was constructed with the current reconciliation `ctx` and utilized at token read operations (like Secret reads from kube api etc). This worked for non-caching, because the token retriever was created fresh at each reconcile and was only bound to the reconcile. Now the object itself is cached, and cannot carry a reconciliation-bound ctx.

This was reworked to:
- The **eager fingerprint read**, which happens synchronously inside the reconcile, is
  bound to the live reconcile context instead (`fingerprintIdentityToken(ctx, r)` +
  `withContext`, returning a `boundIdentityTokenRetriever`). A hung API server can no longer
  block a reconcile worker uninterruptibly.
- The **detached read** gets a hard `webIdentityTokenReadTimeout` of 30s. The SDK refreshes
  credentials through `aws.suppressedContext`, which strips deadline and cancellation before
  calling the provider, so nothing upstream can bound that read — a hung token source would
  otherwise pin the refreshing goroutine and its singleflight slot forever.
- The retriever re-reads its token source on every refresh, so a long-lived cache entry
  never pins a token that expires underneath it.

### Credential expiry margin

Assume-role and web identity credentials now carry a 5 minute expiry window with jitter.
Retrieved credentials are handed to the Terraform provider as static values and the API
calls made with them can outlive the reconciliation that retrieved them, so they need a
validity margin; without a window the SDK returns credentials with as little as a second
left. Previously these configs were rebuilt every reconcile, which hid the problem. The
jitter keeps ProviderConfigs minted at the same time from all refreshing at once.

### Observability

Credential cache skip decisions now log a stable reason, and cache lines identify the ProviderConfig by name and UID.

### ProviderConfig e2e test harness/configuration package:

Reworked he `e2e/providerconfig-aws-e2e-test` configuration. It was on Crossplane v1 APIs. It is now brought up to v2, utilizing namespaced XRs. See `e2e/providerconfig-aws-e2e-test/README.md` for usage.

I have:

- added new unit tests
- providerconfig e2e test suite passes

Also manually tested:

- Create the following provider configs:
  - Static creds in `Secret` with AssumeRoleChain
  - IRSA
  - IRSA + AssumeRoleChain
  - WebIdentity
  - WebIdentity + AssumeRoleChain
- Non-cached:
  - PodIdentity
  - Static creds in `Secret`
- Create MRs per ProviderConfig and observe readiness for all.
- Retrigger reconciliations several times manually and observe cache hits for the first group
  - Validate CloudTrail logs for sts calls
- Rotate creds and observe cache miss and fresh STS calls from cloud trail.


erhancagirici:
@jonasz-lasut thanks for the review: rebased and addressed all in https://github.com/crossplane-contrib/provider-upjet-aws/pull/2212/commits/3486b9ca382034e378a0c2a78833943c419a50c0

refactored cache keys to per (ProviderConfig UID, region) tuple
- move (source, generation, credsHash) keys into the entry itself
- cache hit: map slot key hit (pc,region) + entry metadata hit
- replace entry when slot matches but entry does not
        - this effectively evicts superseded credentials for a (pc,region)
        - i.e. A `(pc,region)` has only one active cred cache entry at given time
- effective max cache size is now ProviderConfigs * regions

- Avoid building cache entries (STS calls) in critical section
  - Make cache entries promise-style
  - `singleflight` builds the cache entry async keyed by full cache key.
  - concurrent consumers of the same creds wait for the same creds.

- Added the expiry window for IRSA too.

> Thank you for the contribution @erhancagirici , few small nits:
> 
> * Every rotation mints a new entry and orphans the old one: it holds live credentials, and for WebIdentity also a retriever, an STS client, and a detached context, until LRU eviction at 1000 (creds_cache.go:35, raised from 100). Nothing removes the entry for the same (UID, generation, region, source) with an older fingerprint and I think we should implement an eviction of older fingerprint entries as we may make the cache full really quick with IRSA rotating every ~1h
> * The global write lock is held across STS calls, this is inherited from main, but main only ever missed for IRSA. Now misses happen per source × per ProviderConfig × per region × per rotation and that means that under STS throttling nothing is cached, so every reconcile serializes and we may hit really long reconciliation times.
> * IRSA never gets the expiry window, plain source: IRSA still hands Terraform a credential that may have seconds of life left, only IRSA with an assumeRoleChain is fine, because the chain re-wraps. From a quick check the fix should be a one-liner
> 
> ```go
>   cfg, err := config.LoadDefaultConfig(ctx, userAgentV2,
>       config.WithRegion(region),
>       config.WithCredentialsCacheOptions(withCredsExpiryWindow), <--
>   )
> ```

erhancagirici:

jonasz-lasut (approved):
Hi @erhancagirici thank you for updating the PR, LGTM!

PR #2227, opened 2026-09-11
fix(cloudfront): map origin request policy by id, not name

### LINKED ISSUE #2226: OriginRequestPolicy is mapped as NameAsIdentifier but is imported by id, so it can never be observed
### Affected Resource(s)

- `cloudfront.aws.upbound.io/v1beta1/OriginRequestPolicy`
- `cloudfront.aws.m.upbound.io/v1beta1/OriginRequestPolicy`

### What happened?

`aws_cloudfront_origin_request_policy` is configured as `config.NameAsIdentifier`, but the resource is imported by **id**, not by name. As a result the managed resource can never be observed: every reconcile concludes the policy does not exist and issues a create, which either fails with `OriginRequestPolicyAlreadyExists` or — when the name happens to be free — succeeds and leaves behind an extra policy.

[`config/externalname.go`](https://github.com/crossplane-contrib/provider-upjet-aws/blob/main/config/externalname.go#L644-L645):

```go
// No import documented, but https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_origin_request_policy#name
"aws_cloudfront_origin_request_policy": config.NameAsIdentifier,
```

The comment reflects the assumption behind the mapping, but import *is* documented, and it is by id:

```console
% terraform import aws_cloudfront_origin_request_policy.policy ccca32ef-dce3-4df3-80df-1bd3000bc4d3
```

<https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_origin_request_policy#import>

Why this cannot work, concretely. `NameAsIdentifier` in upjet is:

```go
NameAsIdentifier = ExternalName{
    SetIdentifierArgumentFn: func(base map[string]any, name string) { base["name"] = name },
    GetExternalNameFn: IDAsExternalName,
    GetIDFn:           ExternalNameAsID,
    OmittedFields:     []string{"name", "name_prefix"},
}
```

- `OmittedFields` removes `name` from `spec.forProvider`, so the only place to supply the policy name is `crossplane.io/external-name`.
- `GetIDFn: ExternalNameAsID` then uses that same value as the Terraform id.
- The CloudFront API reads an origin request policy by id (a UUID). A name is never a valid id, so `Observe` always reports the resource as absent.

Every sibling CloudFront policy resource in the same file is already `IdentifierFromProvider`:

```go
"aws_cloudfront_cache_policy":            config.IdentifierFromProvider,
"aws_cloudfront_response_headers_policy": config.IdentifierFromProvider,
"aws_cloudfront_origin_access_control":   config.IdentifierFromProvider,
```

`ResponseHeadersPolicy` is the useful contrast: it keeps `name` in `spec.forProvider`, takes a provider-assigned id as its external name, and works. `OriginRequestPolicy` differs only in this one config entry.

### Steps to Reproduce

1. Create an `OriginRequestPolicy` with `crossplane.io/external-name: ExamplePolicy` (the only way to set the policy name, since `name` is omitted from the CRD).
2. Let it reconcile once. It creates a policy named `ExamplePolicy`; the provider writes the new UUID back into the external-name annotation.
3. Reset the external name to `ExamplePolicy` — this happens on every reconcile if a Composition sets the annotation, which it must in order to control the policy name.
4. `Observe` looks the policy up by id `ExamplePolicy`, does not find it, and calls `CreateOriginRequestPolicy` again.

From then on the resource never reaches `Ready`. If the name is taken the create 409s; if the external name currently holds a UUID, the create succeeds and produces a policy **named after a UUID**, which is then orphaned on the next pass.

```yaml
apiVersion: cloudfront.aws.upbound.io/v1beta1
kind: OriginRequestPolicy
metadata:
  name: example
  annotations:
    crossplane.io/external-name: ExamplePolicy
spec:
  forProvider:
    comment: Example origin request policy
    cookiesConfig:
      cookieBehavior: none
    headersConfig:
      headerBehavior: whitelist
      headers:
        items:
          - User-Agent
    queryStringsConfig:
      queryStringBehavior: none
```

```
create failed: async create failed: failed to create the resource:
[{0 creating CloudFront Origin Request Policy (ExamplePolicy): operation error CloudFront:
CreateOriginRequestPolicy, https response error StatusCode: 409,
OriginRequestPolicyAlreadyExists: Another origin request policy with the same name
already exists within the aws account.  []}]
```

Once the resource has been through a failed create it also tends to land on:

```
cannot determine creation result - remove the crossplane.io/external-create-pending
annotation if it is safe to proceed
```

### Proposed fix

```diff
-	// No import documented, but https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_origin_request_policy#name
-	"aws_cloudfront_origin_request_policy": config.NameAsIdentifier,
+	// Cloudfront Origin Request Policies can be imported using the id
+	"aws_cloudfront_origin_request_policy": config.IdentifierFromProvider,
```

This also restores `name` and `name_prefix` to `spec.forProvider`, so the resource ends up behaving exactly like `ResponseHeadersPolicy`.

One caveat to flag: this changes external-name semantics for existing users. The external name becomes the policy id rather than its name, and `name` moves into `spec.forProvider`, so existing resources need their external name set to the policy id to be adopted rather than recreated.

PR: #2227. It carries the config change on its own so the direction can be settled before the generated diff lands on top.

### Crossplane Version

v2.1.4

### Provider Version

provider-aws-cloudfront v2.5.3

### Kubernetes Version

v1.36

### Kubernetes Distribution

EKS

ottramst:

`aws_cloudfront_origin_request_policy` is mapped as `NameAsIdentifier`, but Terraform imports this resource by id. `NameAsIdentifier` drops `name` from the CRD and uses the external name as the Terraform id, so the policy can only ever be addressed by a name that the CloudFront API cannot look up. The resource is therefore never observed: every reconcile tries to create it, which 409s once the name exists and otherwise leaves a stray policy behind.

Switching it to `IdentifierFromProvider` matches `aws_cloudfront_cache_policy`, `aws_cloudfront_response_headers_policy` and `aws_cloudfront_origin_access_control`, which all already work this way.

Full analysis, reproduction and error output are in the issue.

Fixes #2226

Two commits: the config change, then `make generate`. The generated diff is small and entirely confined to this resource — `name` and `name_prefix` return to `spec.forProvider`, the CRDs gain a CEL rule requiring `name` when Create or Update is in `managementPolicies`, and `managed.NewNameAsExternalName` drops out of both controllers.

### One thing I would like input on

This changes external-name semantics for existing users. The external name becomes the policy id rather than its name, so existing resources need theirs updated or they will be seen as absent and recreated. I am glad to add a migration note wherever you would like it, or to approach the transition differently if you would prefer.

`make generate` and `make check-diff` both run clean locally (`branch is clean`).

The bug was reproduced on provider-aws-cloudfront v2.5.3 with Crossplane v2.1.4: an `OriginRequestPolicy` whose policy already existed in the account never reached `Ready`, retrying `CreateOriginRequestPolicy` and getting `OriginRequestPolicyAlreadyExists` each time. Setting the external name to the policy id made it reconcile immediately, which is the behaviour this change makes the default.

Not yet exercised through uptest.

I have:

### DIFF
--- a/examples-generated/cluster/cloudfront/v1beta2/originrequestpolicy.yaml
+++ b/examples-generated/cluster/cloudfront/v1beta2/originrequestpolicy.yaml
@@ -24,6 +24,7 @@ spec:
       headers:
         items:
         - example
+    name: example-policy
     queryStringsConfig:
       queryStringBehavior: whitelist
       queryStrings:
--- a/apis/cluster/cloudfront/v1beta2/zz_generated.deepcopy.go
+++ b/apis/cluster/cloudfront/v1beta2/zz_generated.deepcopy.go
@@ -6851,6 +6851,11 @@ func (in *OriginRequestPolicyInitParameters) DeepCopyInto(out *OriginRequestPoli
 		*out = new(OriginRequestPolicyHeadersConfigInitParameters)
 		(*in).DeepCopyInto(*out)
 	}
+	if in.Name != nil {
+		in, out := &in.Name, &out.Name
+		*out = new(string)
+		**out = **in
+	}
 	if in.QueryStringsConfig != nil {
 		in, out := &in.QueryStringsConfig, &out.QueryStringsConfig
 		*out = new(OriginRequestPolicyQueryStringsConfigInitParameters)
@@ -6933,6 +6938,11 @@ func (in *OriginRequestPolicyObservation) DeepCopyInto(out *OriginRequestPolicyO
 		*out = new(string)
 		**out = **in
 	}
+	if in.Name != nil {
+		in, out := &in.Name, &out.Name
+		*out = new(string)
+		**out = **in
+	}
 	if in.QueryStringsConfig != nil {
 		in, out := &in.QueryStringsConfig, &out.QueryStringsConfig
 		*out = new(OriginRequestPolicyQueryStringsConfigObservation)
@@ -6967,6 +6977,11 @@ func (in *OriginRequestPolicyParameters) DeepCopyInto(out *OriginRequestPolicyPa
 		in, out := &in.HeadersConfig, &out.HeadersConfig
 		*out = new(OriginRequestPolicyHeadersConfigParameters)
 		(*in).DeepCopyInto(*out)
+	}
+	if in.Name != nil {
+		in, out := &in.Name, &out.Name
+		*out = new(string)
+		**out = **in
 	}
 	if in.QueryStringsConfig != nil {
 		in, out := &in.QueryStringsConfig, &out.QueryStringsConfig
--- a/apis/cluster/cloudfront/v1beta2/zz_originrequestpolicy_types.go
+++ b/apis/cluster/cloudfront/v1beta2/zz_originrequestpolicy_types.go
@@ -104,6 +104,9 @@ type OriginRequestPolicyInitParameters struct {
 	// Object that determines whether any HTTP headers (and if so, which headers) are included in the origin request key and automatically included in requests that CloudFront sends to the origin. See Headers Config for more information.
 	HeadersConfig *OriginRequestPolicyHeadersConfigInitParameters `json:"headersConfig,omitempty" tf:"headers_config,omitempty"`
 
+	// Unique name to identify the origin request policy.
+	Name *string `json:"name,omitempty" tf:"name,omitempty"`
+
 	// Object that determines whether any URL query strings in viewer requests (and if so, which query strings) are included in the origin request key and automatically included in requests that CloudFront sends to the origin. See Query String Config for more information.
 	QueryStringsConfig *OriginRequestPolicyQueryStringsConfigInitParameters `json:"queryStringsConfig,omitempty" tf:"query_strings_config,omitempty"`
 }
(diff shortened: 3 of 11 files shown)


examples-generated/cluster/cloudfront/v1beta2/originrequestpolicy.yaml
```diff
@@ -24,6 +24,7 @@ spec:
       headers:
         items:
         - example
+    name: example-policy
```
jonasz-lasut: Could you add `spec.forProvider.name` to `examples/`? Both cluster and namespaced scoped

ottramst: Done in dd7732bfb8: added `spec.forProvider.name` to the cluster (v1beta1, v1beta2) and namespaced examples.

jonasz-lasut:
Could you sign-off DCO? I'll run ci and e2e tests after

ottramst:
DCO done

jonasz-lasut:
/test-examples="examples/cloudfront/cluster/v1beta2/originrequestpolicy.yaml"

jonasz-lasut:
/test-examples="examples/cloudfront/cluster/v1beta2/originrequestpolicy.yaml"

jonasz-lasut (approved):
Thank you for your contribution @ottramst LGTM!

PR #2238, opened 2026-09-17
fix: honor skip_requesting_account_id in Terraform client setup

### LINKED ISSUE #2213: [Bug]:  S3 Bucket observe fails against custom S3 endpoints (S3 Control tagging fallback misses dial/DNS errors)

- [x] I have searched the existing issues

### Affected Resource(s)

`s3.aws.m.upbound.io/v1beta1, Kind=Bucket` (any `Bucket` reconciled against a custom, non-AWS S3 endpoint. Reproduced on LocalStack; possible on MinIO, Cloudflare R2, OVHcloud)

```yaml
apiVersion: aws.m.upbound.io/v1beta1
kind: ClusterProviderConfig
metadata:
  name: default
spec:
  credentials:
    source: Secret
    secretRef: { name: aws-localstack-credentials, namespace: localstack, key: creds }
  endpoint:
    source: Custom
    hostnameImmutable: true
    url:
      type: Static
      static: http://localstack.localstack.svc.cluster.local:4566
    services: [s3, s3control, kms, rds, elasticache, sts]  # trimmed
  s3_use_path_style: true
  skip_credentials_validation: true
  skip_metadata_api_check: true
  skip_region_validation: true
  skip_requesting_account_id: true
---
apiVersion: s3.aws.m.upbound.io/v1beta1
kind: Bucket
metadata:
  name: example
  namespace: default
spec:
  forProvider:
    region: us-east-1
    forceDestroy: true
  providerConfigRef:
    kind: ClusterProviderConfig
    name: default
```

### Steps to Reproduce

1. Run LocalStack in-cluster, reachable at `localstack.localstack.svc.cluster.local:4566`.
2. Apply the `ClusterProviderConfig` and `Bucket` above.
3. Bucket reaches `Ready=True` (it is created), then observe fails on every reconcile.
4. `kubectl describe bucket example` → `Synced=False`.

Note `skip_requesting_account_id: true` is already set — that is central to the diagnosis below.

### What happened?

Starting in **v2.5.0**, every `Bucket` observe fails when the provider points at a custom S3 endpoint. `Synced` goes `False` and stays there. The bucket itself is created fine, and `Ready` stays `True`, so this is easy to miss; the resource looks healthy while drift detection is dead.

The provider calls S3 Control's `ListTagsForResource` on every observe. The AWS SDK addresses S3 Control at `<account-id>.<endpoint-host>`, so the request goes to:

```
http://000000000000.localstack.localstack.svc.cluster.local:4566/v20180820/tags/arn%3Aaws%3As3%3A%3A%3Aexample-6bf68886
```

That hostname has no DNS record: real AWS resolves the account-prefixed form via wildcard DNS, an in-cluster Service does not, so the request fails at **dial time** with `no such host`.

There *is* a fallback to the plain S3 tagging API, and the vendored fork does contain it (`internal/service/s3/tags.go`, `ListTags`, "Bucket" case):

```go
if accountID := c.AccountID(ctx); accountID != "" {
    // Attempt ListTagsForResource first, fall back to GetBucketTagging.
    tags, err = tfs3control.ListTags(ctx, c.S3ControlClient(ctx), bucketARN(ctx, c, identifier), accountID)
    if errs.Contains(err, "is not authorized to perform: s3:ListTagsForResource") ||
        tfawserr.ErrCodeEquals(err, errCodeMethodNotAllowed) {
        tags, err = bucketListTags(ctx, conn, identifier)
    }
} else {
    tags, err = bucketListTags(ctx, conn, identifier)
}
```

Two reasons it never fires here:

1. **The guard is the wrong test.** It keys off `accountID != ""`. With `skip_requesting_account_id: true`, the account ID is not empty — it is the placeholder `000000000000` (LocalStack's default account, and the value the Upbound fork deliberately pads to 12 digits). So the `else` branch — the one that would do the right thing — is unreachable in exactly the configuration it was meant to serve.
2. **The error match is too narrow.** Both conditions expect an *AWS error code* (`not authorized`, `MethodNotAllowed`). A dial/DNS failure never reaches the server, so there is no code to match. The error surfaces as `StatusCode: 0` with an empty `RequestID`.

**Suggested fix:** fall back whenever the S3 Control attempt fails in a way that indicates the endpoint does not implement it, including transport-level errors (`StatusCode: 0` / `net.Error` / `no such host`). Alternatively, gate the S3 Control path on the endpoint being real AWS rather than on `accountID != ""`. A custom `endpoint.url` with `skip_requesting_account_id` is a strong signal that S3 Control is unavailable.

```shell
observe failed: failed to observe the resource: [{0 listing tags for S3 (Simple Storage) Bucket (example-6bf68886): operation error S3 Control: ListTagsForResource, https response error StatusCode: 0, RequestID: , HostID: , request send failed, Get "http://000000000000.localstack.localstack.svc.cluster.local:4566/v20180820/tags/arn%3Aaws%3As3%3A%3A
%3Aexample-6bf68886": dial tcp: lookup 000000000000.localstack.localstack.svc.cluster.local on 10.96.0.10:53: no such host []}]
```

### Crossplane Version

2.3.4

### Provider Version

2.5.0+

### Kubernetes Version

v1.35.0

### Kubernetes Distribution

kind

### Additional Info

Each version installed in-cluster with the AWS family pinned to match, default cluster DNS, forced fresh observe:

| provider-aws-s3 | vendored fork commit | Result |
|---|---|---|
| v2.4.0 | `2025-12-04` | **works** (`Synced=True`) |
| v2.5.0 | `2026-03-05` | fails |
| v2.5.5 | — | fails |
| v2.6.3 | — | fails |
| v2.7.1 | — | fails |

First bad version: **v2.5.0**. This pins us to v2.4.0, which blocks all AWS provider upgrades, since the family must share a minor line.

Ruled out as workarounds (all tested):

- `endpoint.hostnameImmutable` / `source: Custom` — act above the layer that synthesises the host
- `s3_use_path_style` — S3 only; the Go SDK has no s3control equivalent
- `AWS_ACCOUNT_ID_ENDPOINT_MODE=disabled` — governs S3 account-ID endpoints, not S3 Control
  (verified via the provider pod and reproduced independently in botocore `--debug`)
- Removing `tags` — the call is unconditional; an untagged Bucket fails identically
- `bucketNamespace: account-regional` — requires S3 directory-bucket naming, fails validation
- Dropping `s3control` from `endpoint.services` — sends the request to **real AWS** (worse)

TerjeLafton:

This PR honors `spec.skip_requesting_account_id` when configuring the embedded Terraform AWS client.

`provider-upjet-aws` currently always injects an account ID into that client. When credential validation is skipped, this is the synthetic `000000000000` value. This PR leaves the embedded Terraform client's account ID empty when `skip_requesting_account_id: true` is explicitly configured.

This relates to #2213. Recent Terraform AWS provider changes select the S3 Control tagging API for ordinary buckets when an account ID is present. S3-compatible implementations that do not support S3 Control then fail bucket observation, even though standard S3 `GetBucketTagging` works:
https://github.com/upbound/terraform-provider-aws/blob/70894c6370d2000b17550a7f87c0ada3654a9254/internal/service/s3/tags.go#L176-L187

The change is opt-in and preserves the existing behavior unless `skip_requesting_account_id: true` is set.

This is intended to start discussion on the correct compatibility boundary. An alternative may be a new, more specific ProviderConfig option for S3-compatible endpoints.

Fixes #2213 

I have:

I didn't add any yet until we have decided on a solution. Better test coverage should be added to prove the solution should be added afterwards. 


jonasz-lasut:
Hi @TerjeLafton thank you for your contribution! Could you sign DCO?

TerjeLafton:
> Hi @TerjeLafton thank you for your contribution! Could you sign DCO?

I am sorry! It is done now.

jonasz-lasut:
Hi @TerjeLafton thank you! We'll need to review the impact of exposing this flag properly as it changes the behavior of multiple resources

TerjeLafton:
Hey @jonasz-lasut, thanks for the reply!

Yeah, I understand that this could have a fairly broad impact. I also don't know the full history behind why we default the account ID this way, so there may be good reasons for preserving the current behavior.

Just to reiterate, though, I don't see this as the only possible solution. It was mostly the simplest change I could find to open the discussion. Adding a new, more specific flag to bypass the default account ID might be a safer and more backwards-compatible approach, depending on what other effects that could have for users enabling it.

The Terraform AWS provider itself doesn't appear to require such a default account ID, so it should be possible to avoid setting one when appropriate.

Happy to make changes or investigate further if there's anything I can do to help move this forward.

